PR #3, opened 2018-04-30
serialize Id as string and use pipe as delimiter

reece:
*WARNING* This PR changes nearly all computed identifiers

* serialize Id as string, not as `<Identifier…>`
* use pipe as instance variable separator

This change was motivated by a good catch by @srynobio that Id objects (strings) were essentially being parsed into Identifier objects and serialized as `<Identifier:VMC:...>` rather than as strings. There's no reason to be doing that. This PR changes the serialization to inline the Id (string) verbatim. Because this string will contain a colon in the typical case, also switch the delimiter from colon (:) to pipe (|). (This wasn't strictly necessary, but if we're invalidating old digests, might as well remove the delimiter conflict too.)


reece:
@srynobio I'd appreciate your comments on this PR. Do you agree with the motivation?

larrybabb:
@reece @srynobio While you are making this change, please consider using a delimiter other than the pipe since these digests may presumably be used in HL7 v2 which uses pipes, carrets and tildes as default delimiters for messages, records, fields, etc..  

Not a huge deal, but if it can be avoided it would be nice.  HL7 v2 has ways to deal with escaping delimiter values in data fields and identifiers, its just that lots of folks might use it to resist implementing. (possibly)

reece:
I think we have a misunderstanding.

The separator is an internal implementation detail that absolutely will not
affect HL7's use of VMC digests. The digest structure and character set is
unchanged with this PR.

There is the potential for a problem if VMC serialization is used to embed
a VMC object verbatim in an HL7 message. I suspect the right way to do that
is to base64 encode in the HL7 message, which would eliminate the conflict.

-Reece

srynobio:
@reece I understand the changes you've made, and I like it for the same reasons I stated.

reece:
Thanks for the review @srynobio. Merged.

PR #35, opened 2019-12-13
add missing dependency

matthdsm:


reece:
Yikes! Yes, thanks for the fix.
I didn't catch this because I usually install with extras, installs it.

PR #45, opened 2020-12-23
updating notebooks to current version of spec

andreasprlic:
This PR updates the notebooks to version 0.6.0 of the spec. (mostly replacing `to_hgvs()` with `translate_to(allele, "hgvs")`.

Also extended an existing test to do a full "roundtrip" (from and to hgvs) in `test_translator.py`


andreasprlic:
Did not update the `Transcript Location` notebook, since that transcript model in that one is not in the main branch.

reece:
Fabulous! Thanks @andreasprlic !

Issue #28, opened 2019-06-25
Add support for translating dup variants from HGVS


melissacline:
The VR translator currently doesn't support translation of dup variants from hgvs, and throws exceptions upon encountering one.  Sample variants include:
NC_000013.11:g.32316467dup
NC_000013.11:g.32319315dup
NC_000013.11:g.32331093_32331094dup

PR #51, opened 2021-01-23
#28 adds support for hgvs dup translation. 

andreasprlic:
… Also introduces dependency on uta database, since that is needed for the normalizer.

I realized that not being able to work with dup variants, is pretty much a showstopper on our side. As such I am trying to fix this. One of the reasons why I suspect this was not working yet, is since this requires to add a new dependency on the UTA database. I tried to encapsulate this in the new HgvsTools class, but from an application perspective it would be nicer to be able to provide a different connection from the outside. I'd be happy to hear suggestions how to improve this wrapper.


reece:
Hi @andreasprlic. Thanks for the contribution on this languishing issue.

Tests aren't passing, so I didn't spend too much time on this PR. Also, I'm wary to have a UTA dependency right now. If you're interested in pursuing this, let's talk it through.

andreasprlic:
Hi @reece ,

I understand the reluctancy about adding UTA as a dependency. Any other suggestions how to enable hgvs dups here? They  are pretty important to be able to use vrs-python from my perspective. As such yes, I am interested in pursuing this and happy to contribute.

About the failed tests: They pass locally. I think I made the initial fork at a time when the devel branch also had some travis issues. I merged the latest devel branch back in. Now the tests are failing with some semantic versioning issue: ```ValueError: choosing custom numbers for the `.devX` distance is not supported.
 The 0.7.0.dev1 can't be bumped
Please drop the tag or create a new supported one```

Any thoughts on how to resolve?

Thanks,
Andreas

andreasprlic:
FYI, I think GitHub actions could not get run on my branch, until I placed a `0.7.0.rc2` tag on my branch. The previous .devX tag somehow resulted in an unhappy build.

reece:
@andreasprlic : Thanks for the dup support! Nicely done. Tests pass for me.

PR #53, opened 2021-05-04
update VRS and apply VRS patch for pjs

ahwagner:
This addresses ga4gh/vrs#295 in vrs-python.


reece:
@ahwagner : This PR was showing conflicts with main, perhaps as a result of merging the hgvs dup PR. The above series of commits on this branch are an attempt to get it into a mergeable state. Please stand by.

PR #54, opened 2021-05-12
upgrade python-jsonschema-objects and vrs submodule

joesolly:


ahwagner (approved):
Updates to most recent VRS `main` commit and passes tests. Seems reasonable to me.

reece:
This seemingly simple PR created lots of merge problems because of the concurrent change in the vrs submodule. I thought that I had worked out the issue merged this PR on May 12, but was mistaken (and then forgot about it :-()

I manually incorporated the setup.cfg changes, tested, and pushed.

Thanks @joesolly . Apologies for the very tardy followup.

PR #55, opened 2021-06-28
Add rudimentary VCF translation

jsstevenson:
This PR extends the Translation class to provide basic support for importing and exporting between VRS and VCF data. Happy for feedback on anything/everything, but it'd be helpful to hear about these items in particular:
* [Test cases](https://github.com/GenomicMedLab/vrs-python/blob/566ee01e2dd4c7e4945190cb870edca5b90f5b07/tests/extras/test_translator.py#L191): I added an excessive amount to compensate for my lack of confidence about correctly reversing the normalization algorithm. Let me know if anything looks wrong or if there appear to be holes (or if this is too many cases).
* I think the normalized Translator fixture breaks TravisCI - is something I can do (eg check for an env var) to prevent those tests from running in CI?
* Frankly, I don't quite understand the particulars of how the VRS repo works as a submodule here, but there's a diff listed below, so hopefully I didn't do something wrong there

I ran everything through the YAPF config in root before committing, which is why there are a bunch of cosmetic alterations

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -98,6 +98,7 @@ extras =
     biocommons.seqrepo>=0.5.1
     bioutils>=0.5.2
     hgvs>=1.4
+    scikit-allel
     requests
 notebooks =
     ipython
--- a/setup.cfg
+++ b/setup.cfg
@@ -98,7 +98,6 @@ extras =
     biocommons.seqrepo>=0.5.1
     bioutils>=0.5.2
     hgvs>=1.4
-    scikit-allel
     requests
 notebooks =
     ipython
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -286,12 +286,13 @@ class Translator:
             return None
         return model(**var)
 
-    # assembly ID patterns
+    # patterns for from_VCF
     chrom_re = re.compile(r'(chr)?(?P<chrom>[0-2]?[0-9])', re.IGNORECASE)
-    grch_re = re.compile(r'GRCh(?P<version>3[8|9])', re.IGNORECASE)
+    grch_re = re.compile(r'GRCh(?P<version>3[7|8])', re.IGNORECASE)
     ncbi_re = re.compile(r'NCBI(?P<version>3[4|5|6])', re.IGNORECASE)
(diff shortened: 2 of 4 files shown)


larrybabb (approved):
I'll defer to @reece and @ahwagner since I have not had my hands or head in the VRS-python code base much.

reece (changes requested):
Hi @jsstevenson . Thanks for the PR! I apologize for the delay -- I'm on vacation.

I have several general requests and comments:

As is, your changes must not modify submodules/vrs to a prior version of VRS. Please merge main into your branch to minimize changes. (Need help? ping me on slack) 

For code structure, this module is concerned with converting *variants*, not *variant files*. Please move the file conversion to a separate module dedicated to VCF file support. It should probably use the translator code. 

Part of my goal here is to cleave off the pysam library, potentially making it optional. Pysam includes htslib, which is pretty heavy. That can be decided later. Ideally, translator.py will not import pysam. (Caveat: I'm not sure I've thought that through well enough to know the downsides.)

As part of VCF support, please consider rewriting the VCF file with VRS ids in the file. [This code](https://github.com/reece/vcf-digester) might help you get started.

As a reviewer, it's a bit confusing when formatting changes because it's marked as a diff, but requires scrutiny to determine that it's just whitespace. Please minimize that. (The `make reformat` target will reformat to a standardized formatting.)

jsstevenson:
@reece thanks (belatedly on my end) for your feedback/thoughts. I got put a highish priority project right after opening this PR, so I've been behind on responding/updating accordingly (and sorry about the format -- I think I manually ran yapf rather than using the Make target).

One question: avoiding an htslib dependency makes sense to me -- do you know of a lighter VCF library that I could draw on, or do you think it'd be sufficiently straightforward to just write a basic parser? Googling yielded one other major package, [PyVCF](https://pyvcf.readthedocs.io/en/latest/INTRO.html), but I was a little leery (no commits for almost five years, not updated for vcf4.2) and it... also requires a PySAM installation.

reece:
Yes, please stay away from PyVCF because it appears to be unmaintained. I would use pysam for VCF access. VCFpy is also available, but I've not used it.

There are now file conflicts in this PR. Please address those too. (We can't merge when there are conflicts.)

PR #57, opened 2021-08-24
Add support for VRS 1.2 Classes in _normalize_allele

korikuzma:
I am using `_normalize_allele` in our variation normalizer. I get an `AttributeError` when I try to normalize an Allele whose fields are classes found in v1.2.

Example: 
```
allele = {'type': 'Allele',
'location': {'type': 'SequenceLocation',
  'sequence_id': 'refseq:NC_000023.11',
  'interval': {'type': 'SequenceInterval',
   'start': {'type': 'IndefiniteRange', 'value': 31120496, 'comparator': '<='},
   'end': {'type': 'IndefiniteRange', 'value': 33339477, 'comparator': '>='}}},
 'state': {'type': 'LiteralSequenceExpression', 'sequence': ''}}
normalize(models.Allele(**allele), dp)

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-770117f84d20> in <module>
----> 1 normalize(allele, dp)
~/.local/share/virtualenvs/variation-normalization-BEk10_rZ/lib/python3.9/site-packages/ga4gh/vrs/normalize.py in normalize(vo, data_proxy)
     72     if vo_type in handlers:
     73         handler = handlers[vo_type]
---> 74         return handler(vo, data_proxy)
     75 
     76     # No handler for vo_type; pass-through unchanged
~/.local/share/virtualenvs/variation-normalization-BEk10_rZ/lib/python3.9/site-packages/ga4gh/vrs/normalize.py in _normalize_allele(allele, data_proxy)
     20 def _normalize_allele(allele, data_proxy):
     21     sequence = SequenceProxy(data_proxy, allele.location.sequence_id._value)
---> 22     ival = (allele.location.interval.start._value, allele.location.interval.end._value)
     23     try:
     24         alleles = (None, allele.state.sequence._value)
~/.local/share/virtualenvs/variation-normalization-BEk10_rZ/lib/python3.9/site-packages/python_jsonschema_objects/classbuilder.py in __getattr__(self, name)
    257         if name in self._extended_properties:
    258             return self._extended_properties[name]
--> 259         raise AttributeError(
    260             "{0} is not a valid property of {1}".format(name, self.__class__.__name__)
    261         )
AttributeError: _value is not a valid property of IndefiniteRange
```


reece:
Nice work, @korikuzma!. Tests pass for me locally too. 

I'm disappointed by `value`/`_value` shenanigans, but that's a schema and python-jsonschema-objects issue.

src/ga4gh/vrs/normalize.py
```diff
@@ -30,9 +42,13 @@ def _normalize_allele(allele, data_proxy):
                                            alleles=alleles,
                                            mode=NormalizationMode.EXPAND,
                                            anchor_length=0)
-        new_allele.location.interval.start = new_ival[0]
-        new_allele.location.interval.end = new_ival[1]
-        new_allele.state.sequence = new_alleles[1]
+        try:
+            new_allele.location.interval.start = new_ival[0]
+            new_allele.location.interval.end = new_ival[1]
+            new_allele.state.sequence = new_alleles[1]
```
reece, line 48: In the above, sequence from an RSE will incorrectly attempt to update `state.sequence`.

I think section would be made clearer by handling the states with sequence and RSE cases completely separately.

Is there a test for the RSE case?

reece (changes requested):
Thanks, @korikuzma!

A few requests:

1) See comments in normalize for what I think is a bug when normalizing an RSE.
2) Please add a test case for RSE normalization.
3) Supporting SimpleInterval and SequenceInterval concurrently with the `value`/`_value` shenanigans is ugly. SimpleInterval is deprecated. To make the intention clearer, please switch on object.type explicitly (rather than implicitly by exception). This will also help us get rid of the SimpleInterval code in 2.0.

Nice work!

src/ga4gh/vrs/normalize.py
korikuzma: Hi @reece , 

I have pushed my changes for separating out interval.type and state.type. 

If RSE is RepeatedSequenceExpression, I did make a test case for this. [Here](https://github.com/GenomicMedLab/vrs-python/blob/vrs-1.2.0/tests/test_vrs_normalize.py#L92) is the allele that has a RSE as the state. 

Please let me know if these changes addressed your suggestions or if I should make additional changes.

reece: Yep, great! Thanks for rounding out support for these cases.

notebooks/Untitled.ipynb
```diff
@@ -0,0 +1,519 @@
+{
```
reece, line 1: @korikuzma : Was this Untitled notebook added by mistake? Please either delete it or rename it.

reece (changes requested):
Nearly there. Please see comment re: Untitled notebook. I suspect that you didn't intend to add that to the PR.

notebooks/Untitled.ipynb
korikuzma: Ahhh my bad. Removed it!

reece (approved):
Yay!

Issue #58, opened 2021-09-03
Would be nice to have a jupyter notebook to provide CNV examples


andreasprlic:
To explain better how to represent CNVs, can we get a jupyter notebook?

andreasprlic:
I can't assign this to myself somehow, but I will provide a PR.

ahwagner:
closed in #59.

PR #59, opened 2021-09-03
addressing #58 by adding a new jupyter notebook for CNV variants.

andreasprlic:
This PR provides an initial example for how to represent copy number variants with VRS. It provides two examples, one of a copy number gain where it is uncertain if the even is in tandem. The second example describes an event where it has been observed that the duplication is in tandem.

What is still missing from this notebook is an example that demonstrates how to work with precise/imprecise coordinate endpoints. What model would I use for that?

Also, please note the example with `RepeatedSequenceExpression`.  It appears that in the current version of the spec it it necessary to provide `IndefiniteRange` at two different locations here. That seems like a minor issue that would be worth clarification though.


reece:
@andreasprlic : Thanks for this. Nicely written!

The MME example is not what we intended and suggests that we need to document more carefully.

We intended that variation would be reported as either Molecular Variation (on a sequence) or Systemic (somewhere in the cell). Furthermore, Systemic Variation can be *derived from* Molecular Variation. In the MME case a 4-copy SV would be derived from a 4-repeat MV.

Would you please update this example to build a MV using an Allele with an RSE, then show a derived SV?

Does that make sense?

@ahwagner

andreasprlic:
Good points. When using an allele to represent [the MME example](https://github.com/ga4gh/vrs-python/blob/e126c49a345a482fdfed73f626c1da40ca952752/notebooks/images/MME_exon_tandem_dup.png), how to specify the uncertainty around the breakpoint?. The precise breakpoint has not been sequenced since it is intronic.

ahwagner:
I took a look through this notebook and agree with @reece – well written!

As @reece mentioned, when trying to indicate the number of tandem repeats on a molecule, it is a repeated sequence expression Allele. @andreasprlic would you help me understand the situation better? You asked about specifying uncertainty around the breakpoint in the MME example, but I don't follow how one could report this is a tandem duplication event if the precise breakpoints were not sequenced. If so how would it known there is not non-repeating sequence between repeats?

andreasprlic:
@ahwagner good point and you are totally right. If we know this is in tandem, we would have precise breakpoints, since they got sequenced and therefore we know they are in tandem for certain. I will update the example accordingly.

andreasprlic:
I merged main back into this branch and that fixed the failing unit tests. Also updated the `HGVS Translation` runbook which had an error message in it. Can we merge this now?

ahwagner:
Looks good to me. Thanks @andreasprlic. Merging.

ahwagner (approved):

PR #64, opened 2021-09-27
63 seqrepo dp

### LINKED ISSUE #63: SeqRepoDataProxy.get_sequence returns different data when qualified or not
`SeqRepoDataProxy.get_sequence` only works when namespace prefix is included. `SeqRepoRESTDataProxy.get_sequence` works when namespace prefix is included or excluded. See below for an example of this observation. Both dataproxies should work for both inputs. 
Screen Shot 2021-09-27 at 8 42 46 AM

korikuzma:
Closes #63 

~~CI/CD is only failing on Python 3.9 due to issues with installing pysam~~


reece:
Hi @korikuzma : Have you investigated the cause for pysam failures in 3.9? I'm reluctant to merge without understanding that because tests will be broken, which in turn means that we can't make any releases.

korikuzma:
> Hi @korikuzma : Have you investigated the cause for pysam failures in 3.9? I'm reluctant to merge without understanding that because tests will be broken, which in turn means that we can't make any releases.

pysam released a wheel for 3.9 last month. So all tests are now passing 😄

reece (approved):
Looks good to me. Merging.

PR #74, opened 2021-12-22
Add error handling for HGVSDataNotAvailableError during HGVS translation

### LINKED ISSUE #73: Some transcripts raising HGVSDataNotAvailableError during translation from allele -> HGVS
I'm attempting to translate from an Allele representation back to a transcript representation and am running receiving errors like `HGVSDataNotAvailableError: No mapping info available for XM_017001344.1` for a few of the transcripts. I believe this is because of the way we loop through all of the aliases when translating to HGVS. 

https://github.com/ga4gh/vrs-python/blob/4978e88396c8a718576ade805c8b1332b7108496/src/ga4gh/vrs/extras/translator.py#L381-L395

The specific case for `XM_017001344.1` is hit when I have an allele using `NM_001331029.1`. The `XM_017001344.1` alias is returned as well as the expected `NM_001331029.1` by `get_metadata` for the `refseq` namespace. 

```
data_proxy.get_metadata("NM_001331029.1")

{'length': 11291,
 'alphabet': 'ACGT',
 'added': '2016-08-24T05:03:40Z',
 'aliases': ['MD5:a8d7243ffeea4ade1dbc33613887693b',
  'NCBI:NM_001331029.1',
  'refseq:NM_001331029.1',
  'NCBI:XM_017001344.1',
  'refseq:XM_017001344.1',
  ...]}
```

I was thinking one possible solution is to wrap the normalization in a try/except block. Then, if an HGVSDataNotAvailableError is raised, we can catch it and log a warning, without stopping execution flow. This way, the other aliases will still return an HGVS string value.

I am happy to put in a PR, but what do you think of the approach?

sawmonaco:
Resolves #73 

I chose to validate the four types of prefixes that we allow in `def ir_stype(a):`, NM, NP, NG, and NC.

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -386,15 +386,22 @@ class Translator:
             # TODO: use default_assembly_name here
             if ns.startswith("GRC") and namespace is None:
                 continue
+
+            if not (any(a.startswith(pfx) for pfx in ("NM", "NP", "NC", "NG"))):
+                continue
+
             var.ac = a
 
-            if not namespace.startswith("GRC"):
-                # if the namespace is GRC, can't normalize, since hgvs can't deal with it
-                hgvs_tools = self._get_hgvs_tools()
-                parsed = hgvs_tools.parse(str(var))
-                var = hgvs_tools.normalize(parsed)
+            try:
+                if not namespace.startswith("GRC"):
+                    # if the namespace is GRC, can't normalize, since hgvs can't deal with it
+                    hgvs_tools = self._get_hgvs_tools()
+                    parsed = hgvs_tools.parse(str(var))
+                    var = hgvs_tools.normalize(parsed)
 
-            hgvs_exprs += [str(var)]
+                hgvs_exprs += [str(var)]
+            except hgvs.exceptions.HGVSDataNotAvailableError:
+                _logger.warning(f"No data found for accession {a}")
 
         return list(set(hgvs_exprs))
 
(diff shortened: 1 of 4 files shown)


reece:
Hi Sarah. I'm in vacation without a laptop. I'll look next Tuesday. Thanks for the PR!

sawmonaco:
Hey @reece, did you get the chance to look at this PR?

reece:
Hi @sawmonaco. I'm getting a test failure in your PR and also in main. I haven't had a chance to hunt it down yet. Can you please confirm that tests on main and this PR pass for you?

sawmonaco:
Yes, all of the tests in `tests/extras/test_translator.py` pass for me. I get one test error in `tests/extras/test_dataproxy.py`, but I think that's due to something not being configured (`sqlite3.OperationalError: unable to open database file`)

reece:
Yep, that's the problem for me. It's not related to anything you did, but I
would like to fix it before merging PRs.

I have a busy week. I won't get to this until the weekend, probably.

-Reece

korikuzma:
Hi @reece and @sawmonaco . I was able to get tests to pass by changing the `SEQREPO_ROOT_DIR` in the `Makefile` to `SEQREPO_ROOT_DIR=tests/data/seqrepo/latest`. Let me know if that change also fixes things on your end.

sawmonaco:
@korikuzma, yes that fixed it for me as well

ahwagner:
These tests look correct to me. Thanks for the addition, @sawmonaco!

PR #76, opened 2022-03-01
Update SEQREPO_ROOT_DIR in Makefile to fix failing test

korikuzma:
This was commented in #74 . Unsure if we wanted to make this change there or in a new PR. Since no change has been made, I decided just to create a new PR. I know @ahwagner is going to make a new release soon for our team's needs, so it would also be nice to ensure all tests pass before doing so.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ PKGD=$(subst .,/,${PKG})
 PYV:=3.9
 VEDIR=venv/${PYV}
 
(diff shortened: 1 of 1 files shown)


larrybabb (approved):
seems pretty straightforward to me. I defer to Reece and Alex, since I have not been involved with the coding to date.
@ahwagner should we be pulling @andreasprlic in on these PR reviews. I would think he'd be familiar with vrs-python libs potentially?

ahwagner (approved):

PR #79, opened 2022-03-08
Add release workflow action

jsstevenson:
Creates PyPI releases from tags added to commits on the main branch.

* GitHub syntax filters are a little more constrained than normal regex so it's difficult to enforce the [entire PEP440 versioning syntax](https://stackoverflow.com/a/38020327) -- the provided filter just does a pretty lax check for tags that look like semantic versioning
* Someone (@ahwagner) will need to supply PyPI API tokens as GitHub secrets named PYPI_API_TOKEN and TEST_PYPI_API_TOKEN

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,49 @@
+name: Upload tagged commit to PyPI
+on:
+  push:
+    tags:
+    - '**'
+jobs:
+  get_branch:
+    runs-on: ubuntu-latest
+    outputs:
+      branch_name: ${{ steps.get_branch_name.outputs.name }}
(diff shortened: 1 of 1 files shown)


ahwagner:
@reece would you kindly generate the above API tokens for the `GA4GH` account on PyPI? I am assuming you have the login credentials for that. Better to use `GA4GH` than my personal account for the tokens.

reece:
@jsstevenson: Thanks for doing this!

I just invited you to ga4gh.vrs on pypi. You should be able to generate an API key at https://pypi.org/manage/project/ga4gh-vrs/settings/.

Issue #72, opened 2021-11-29
Translate deprecated to non-deprecated classes


ahwagner:
We should enable vrs-python to automatically translate deprecated classes to preferred ones: 

- SimpleInterval -> SequenceInterval
- SequenceState -> LiteralSequenceExpression

korikuzma:
This was added in #86 , and #87

PR #86, opened 2022-04-05
fix: support `SequenceState, SimpleInterval`

aguliyev:
fix: support `SequenceState` with `LiteralSequenceExpression` and `SimpleInterval` with `SequenceInterval`

This PR is main->main

There is similar patch on tag 0.7.4 :
https://github.com/ga4gh/vrs-python/compare/0.7.4...aguliyev:SIPAW-284_Add_support?expand=1


aguliyev:
@andreasprlic , my concern is that VDB uses 0.7.4 , not the latest `main`
so applying this PR is fine, but we need to also update 0.7.4. 
Or, do you mean we update in VDB to use the latest code?

andreasprlic:
@aguliyev 0.7.4 is currently the latest stable release. I believe the next to follow would be 0.7.5. We would switch over to that.

korikuzma:
Instead of allowing both should we instead translate deprecated to non-deprecated models as noted in [this](https://github.com/ga4gh/vrs-python/issues/72) issue?

aguliyev:
according to https://github.com/ga4gh/vrs-python#vrs-python-and-vrs-version-correspondence
there should be

> vrs-python 0.7 branch tracks the vrs 1.2 branch.

but checking the list of branches I do not see it

andreasprlic:
@[korikuzma](https://github.com/korikuzma) can you explain what is meant with "translate" between the two models? What should the code do here?

korikuzma:
> @[korikuzma](https://github.com/korikuzma) can you explain what is meant with "translate" between the two models? What should the code do here?

So if the user inputs an Allele using a SequenceState (deprecated), we would convert it into a LiteralSequenceExpression. Similarly, if the user inputs a SimpleInterval (deprecated), we would convert it into a SequenceInterval.

aguliyev:
> should we instead translate deprecated to non-deprecated models

normally yes; But that will lead to different model, hence different generated identifier(hash).
That is the main reason of us making this fix - we want to continue generating allele VRS hashes from the old model.
(the hash is based on the model serialization value)

andreasprlic:
@[korikuzma](https://github.com/korikuzma) what do you think of the new `ensure_allele_is_latest_model`?

korikuzma:
> @[korikuzma](https://github.com/korikuzma) what do you think of the new `ensure_allele_is_latest_model`?

I like it! @aguliyev raises a good point about people still using deprecated classes. My initial thought is to set a bool in the translate to/from methods on whether or not to allow deprecated classes, where the default is set to only allow or translate to the latest. @andreasprlic , what do you think?

src/ga4gh/vrs/extras/translator.py
```diff
@@ -342,10 +361,8 @@ def ir_stype(a):
                 return "g"
             return None
 
-        if (vo.type != "Allele"
-            or vo.location.type != "SequenceLocation"
-            or vo.state.type != "LiteralSequenceExpression"):
-            raise ValueError("_to_hgvs requires a VRS Allele with SequenceLocation and LiteralSequenceExpression")
+        if not self.is_valid_allele(vo):
+            raise ValueError("_to_hgvs requires a VRS Allele with SequenceLocation and LiteralSequenceExpression|SequenceState")
```
korikuzma, line 365:
```suggestion
            raise ValueError("_to_hgvs requires a VRS Allele with SequenceLocation and LiteralSequenceExpression")
```

src/ga4gh/vrs/extras/translator.py
```diff
@@ -431,22 +448,24 @@ def _to_spdi(self, vo, namespace="refseq"):
         is expected to be normalized per VRS spec.
 
         """
-        if (vo.type != "Allele"
-            or vo.location.type != "SequenceLocation"
-            or vo.state.type != "LiteralSequenceExpression"):
-            raise ValueError("_to_hgvs requires a VRS Allele with SequenceLocation and LiteralSequenceExpression")
+        if not self.is_valid_allele(vo):
+            raise ValueError("_to_spdi requires a VRS Allele with SequenceLocation and LiteralSequenceExpression|SequenceState")
```
korikuzma, line 452:
```suggestion
            raise ValueError("_to_spdi requires a VRS Allele with SequenceLocation and LiteralSequenceExpression")
```

korikuzma (commented):
This may change depending on the codeowners' opinions on my [latest comment](https://github.com/ga4gh/vrs-python/pull/86#issuecomment-1090873859).

src/ga4gh/vrs/extras/translator.py
aguliyev: ok

src/ga4gh/vrs/extras/translator.py
aguliyev: ok

andreasprlic (approved):
LGTM

andreasprlic:
I feel the translate_to methods should be generally safe to use update allele models (that have deprecated mapped to current) models. As long as we are not calculating any hashes, there should be no risk here.

aguliyev:
I dont have merge permissions.
Plz, could somebody merge;  also, should we make tag/release? Because we need to use it in our project...

andreasprlic:
Once this is merged, what will it take to get this out in a release?

aguliyev:
another question - is the current `main` which we patch here a production ready?  (checking git tags, looks like it is dev/alpha).
Should we make a branch off tag 0.7.4 and patch that one also?

FYI @andreasprlic

ahwagner:
@aguliyev this was my concern too; as you pointed out we never made a `0.7` branch, and IIRC `main` currently has some `0.8` dev work in it. I plan to review and clean that up tonight.

As for versioning, @jsstevenson added an [auto-deploy action to GH](https://github.com/ga4gh/vrs-python/blob/main/.github/workflows/release.yml). We just have to version tag a commit on `main` and it will build and deploy to PyPI. Any of the VRS devs should be able to do this.

andreasprlic:
@ahwagner could we just create a new branch off the latest 0.7.X release tag, merge this PR into that branch (as well as main), apply a new tag, and this would give us a new release in that series?

ahwagner:
@andreasprlic and @aguliyev this is now live on PyPI as `0.7.5`.

andreasprlic:
Great, thank you @ahwagner !

aguliyev:
thank you guys.
plz, notice that now 0.7.5 becomes newer than 0.8.x
so, 0.7.5 is no longer tied to vrs 1.2 branch (as it is told in README.
Also, do we want to make a release?

PR #89, opened 2022-04-14
Issue 88: Update vrs submodule + remove type from models in translator

### LINKED ISSUE #88: Don't specify type when redundant to instanciated class
It looks as if right now we need to write 

```models.IndefiniteRange(type="IndefiniteRange", value=1, comparator="<=")```

when the following would be much cleaner

```models.IndefiniteRange(value=1, comparator="<=")```

Can we just assume that the type matches the instantiated class, unless specified differently?

korikuzma:
Closes #88 

@andreasprlic It appears that the vrs-python 0.7 branch is working fine for the vrs 1.2 branch. So I assume that #88 was just regarding main?

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -93,14 +93,12 @@ class Translator:
         """
         if allele.state.type == "SequenceState":
             allele.state = models.LiteralSequenceExpression(
-                type="LiteralSequenceExpression",
                 sequence=allele.state.sequence,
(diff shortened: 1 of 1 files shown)


korikuzma:
@andreasprlic Actually, it looks like vrs-python 0.7 branch doesn't have the latest commits from vrs 1.2. I'll work on this in a new issue.

jsstevenson (approved):

ahwagner (approved):

PR #106, opened 2022-08-17
refactor: Metaschema update

korikuzma:
This PR contains changes for updating models in core/vrs/vrsatile. Will need to make an initial dev release to use in our other packages so that we aren't blocked. 

I made some local changes to the submodules (did not push as @ahwagner will likely be making these changes in the appropriate repo). This includes doing `oneOf` for abstract classes rather than setting `$ref` to the abstract class. python jsonschema objects was having issues with this. 

Still left TODO: 
- [ ] Update Variation Normalizer to fix failing tests
- [ ] Update test for VCF Annotator
- [ ] There will probably be more changes to the submodules, so making those appropriate changes 
- [ ] ordering property on schemas
- [x] think on how to fix removing `id` field when calculating digests affects models such as domain entities.


korikuzma:
Closing. Will be working off the 0.9 branch from now on

Issue #121, opened 2022-11-18
VRS / VCF Annotation Tool should allow files from google cloud bucket


korikuzma:
Collaborators would like this feature

korikuzma:
- [ ] Include README

korikuzma:
They want to be able to read from a bucket, then have the output files uploaded to a bucket

korikuzma:
Will not do this on our side

PR #123, opened 2022-11-18
feat: VCFAnnotator accepts both seqrepo dps + clean up

### LINKED ISSUE #122: VCF / VRS Annotation should allow either local or rest seqrepo dp
Currently, it defaults to `SeqRepoDataProxy(SeqRepo("/usr/local/share/seqrepo/latest"))`, however we should allow users to choose between rest or local. We should also allow them to provide the path/url if not using the default values.

korikuzma:
Close #122

Notes:
- Replaces argparse with click
- Allows users to select from either local or REST SeqRepo Data Proxy
- Cleans up the VCFAnnotator code

### DIFF
--- a/src/ga4gh/vrs/extras/vcf_annotation.py
+++ b/src/ga4gh/vrs/extras/vcf_annotation.py
@@ -1,4 +1,9 @@
-"""Module containing tools for annotating VCFs with VRS"""
+"""Module containing tools for annotating VCFs with VRS
+
+Example of how to run from root of vrs-python directory:
+python3 -m src.ga4gh.vrs.extras.vcf_annotation --vcf_in input.vcf.gz \
+    --vcf_out output.vcf.gz --vrs_file vrs_objects.pkl
+"""
 import pickle
 from enum import Enum
 from typing import Dict, List, Optional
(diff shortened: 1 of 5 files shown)


src/ga4gh/vrs/extras/vcf_annotation.py
```diff
@@ -1,122 +1,168 @@
...
+@click.option(
+    "--seqrepo_base_url",
+    required=False,
+    default="http://localhost:5000/seqrepo",
+    help="The base url for SeqRepo REST API"
+)
+def annotate_click(vcf_in: str, vcf_out: str, vrs_file: str,
+                   seqrepo_dp_type: SeqRepoProxyType, seqrepo_root_dir: str,
+                   seqrepo_base_url: str) -> None:
+    """Annotate VCF file via click"""
```
jsstevenson, line 64: I think retaining the example command (`ex. python3 src/ga4gh/vrs/extras/vcf_annotation.py input.vcf.gz --out ./output.vcf.gz --vrs-file ./vrs_objects.pkl` from the previous commit) might be worthwhile -- you add a new line in here

korikuzma: Probably would be good to include a README for the tool once it’s complete. Can do in #121

jsstevenson: Is that a findable location? I don't know if I'd expect to find an execution example in the `if __name__...` clause. If you put it in the docstring, then it's accessible via the `--help` message, but readme makes sense as well

korikuzma: Ended updating in this branch since #121 will not be addressed in the project.

wesleygoar (approved):

ahwagner (approved):

PR #148, opened 2023-01-13
feat: Add validation to _from_gnomad in translator

### LINKED ISSUE #147: Add input validation for _from_gnomad
I made a [similar issue](https://github.com/ga4gh/vrs-hackathons/issues/7) for translator validation in our hackathon repo. 

`7-7-helloworld-T` is not a valid gnomAD VCF, but we're still able to translate to a VRS Allele as seen below:

Screen Shot 2023-01-13 at 9 23 12 AM

We should add some additional validation checks for the input

korikuzma:
Close #147 

- Update regex for input str
    - Based off of the [The Variant Call Format Specification](https://samtools.github.io/hts-specs/VCFv4.3.pdf) that @wesleygoar sent me:
    Screen Shot 2023-01-13 at 10 56 12 AM
- Validate that expected reference sequence matches actual reference sequence
- Cassettes were automatically generated
- Updated translator test to match latest [VRS docs example](https://vrs.ga4gh.org/en/latest/impl-guide/example.html) since the previous test had an invalid reference sequence

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -4,7 +4,7 @@ Input formats: VRS (serialized), hgvs, spdi, gnomad (vcf), beacon
 Output formats: VRS (serialized), hgvs, spdi, gnomad (vcf)
 
 """
-
+from typing import Tuple
 from collections.abc import Mapping
 import logging
 import re
@@ -23,6 +23,12 @@ from ga4gh.vrs.extras.decorators import lazy_property  # this should be relocate
 from ga4gh.vrs.utils.hgvs_tools import HgvsTools
 
 _logger = logging.getLogger(__name__)
+
+
+class ValidationError(Exception):
+    """Class for validation errors during translation"""
+
+    pass
 
 
 class Translator:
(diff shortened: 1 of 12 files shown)


wesleygoar (approved):

korikuzma:
See this comment regarding requiring valid input or not: https://github.com/ga4gh/vrs-python/issues/147#issuecomment-1384208673

korikuzma:
I decided to raise a `ValueError` if `required_validation=True` since these lines in `translate_from` would not give a descriptive error message
https://github.com/ga4gh/vrs-python/blob/7cd0c830a7bd2fa624063cce8e3fd90fb9cf89cb/src/ga4gh/vrs/extras/translator.py#L81-L82

Open to other ways of implementing!

src/ga4gh/vrs/extras/translator.py
```diff
@@ -56,24 +62,28 @@ def __init__(self,
...
+            fail. If `False` then VRS object will be returned even if validation checks
+            fail.
 
         See also notes about `from_` and `to_` methods.
         """
 
         if fmt:
             t = self.from_translators[fmt]
-            o = t(self, var)
+            o = t(self, var, require_validation=require_validation)
```
jsstevenson, line 80: The dispatch table approach is pretty clean when each translator has the same arguments, but if we need different params for different translators, I think we probably have to call them within some if/then blocks. A little inelegant but IMO cleaner than passing a bunch of unused arguments to every other function.

src/ga4gh/vrs/extras/translator.py
```diff
@@ -175,11 +189,16 @@ def _from_gnomad(self, gnomad_expr, assembly_name=None):
...
-        alt = g["alt"]
+        ref = g["ref"].upper()
+        alt = g["alt"].upper()
         end = start + len(ref)
         ins_seq = alt
 
+        # validation checks
+        valid_ref_seq, err_msg = self._is_valid_ref_seq(sequence_id, start, end, ref)
+        if not valid_ref_seq and require_validation:
+            raise ValidationError(err_msg)
```
jsstevenson, line 200: Do we want to enter this section at all if `require_validation` is false?

jsstevenson (changes requested):

src/ga4gh/vrs/extras/translator.py
korikuzma: I agree. I believe we will be adding validation checks to other methods as well (just haven't gotten around to it). Can make a separate issue for this

src/ga4gh/vrs/extras/translator.py
korikuzma: I'll switch the order!

jsstevenson (approved):

PR #149, opened 2023-01-16
cleanup: resolve pylint warnings

### LINKED ISSUE #97: Resolve pylint errors

korikuzma:
Close #97 

Notes:
- Resolved easy pylint warnings. Can see the warnings in [this](https://github.com/ga4gh/vrs-python/actions/runs/3943712050/jobs/6748860291) action, under `Pylint static analysis`
- For those that had `too-many` warnings, I just disabled. We can fix this in a later issue if we want to do more code clean up.  
- `disable=unused-argument` in translate_from methods will be fixed in #150 
- Disable `fixme` for pylint. It seemed silly to comment it out with a disable
- Renamed `normalize` to `do_normalize` in `Translator` due to `Redefining name 'normalize' from outer scope (line 21) (redefined-outer-name)`. We can rename to something else if we don't like `do_normalize`. I am bad at naming. 
- `localize_allele` was calling a method that did not exist (`localize`), so used `localize_named_feature` which seemed like what it should be calling. 
- `_get_coords` used `chr_cb_map` but the docs said "return (start,end) of band `cb` in map `m`", so switched to using `m`.
- Renamed some variables due to `redefined-outer-name`. Open to better var names. (naming is hard x2)


korikuzma:
@ahwagner I think it'd be good to get your review on this

src/ga4gh/vrs/extras/localizer.py
```diff
@@ -45,12 +42,15 @@ def __init__(self):
...
 
-    def localize_allele(self, allele):
-        # copy input variant and replace location
-        # N.B. deepcopy leads to recursion errors
+    def localize_allele(self, allele, assembly_name = "GRCh38"):
+        """copy input variant and replace location
+        N.B. deepcopy leads to recursion errors
+        """
+        #
+        #
```
jsstevenson, line 50: are these some pylint magic or leftovers from previous commit?

src/ga4gh/vrs/extras/translator.py
```diff
@@ -28,7 +28,7 @@
 class ValidationError(Exception):
     """Class for validation errors during translation"""
 
-    pass
+    pass  # pylint: disable=unnecessary-pass
```
jsstevenson, line 31: you actually don't need this line at all (idk if Pylint will accept that though?)

src/ga4gh/vrs/extras/vcf_annotation.py
```diff
@@ -27,7 +27,7 @@
 class VCFAnnotatorException(Exception):
     """Custom exceptions for VCF Annotator tool"""
 
-    pass
+    pass  # pylint: disable=unnecessary-pass
```
jsstevenson, line 30: ditto above

jsstevenson (approved):
looks good to me. I made a few notes of possible changes but nothing important. Some of these warnings are probably fine to make QC pass but are things that would be good to refactor later (eg functions that are too long or have too many arguments).

I will also say that the amount of times we have to ignore the protected-access warning suggests that maybe we should just disable it project-wide.

src/ga4gh/vrs/extras/translator.py
korikuzma: Ya, I just do it out of habit. I'll remove and try to get better at not defaulting to pass

korikuzma:
> looks good to me. I made a few notes of possible changes but nothing important. Some of these warnings are probably fine to make QC pass but are things that would be good to refactor later (eg functions that are too long or have too many arguments).
> 
> I will also say that the amount of times we have to ignore the protected-access warning suggests that maybe we should just disable it project-wide.

In my original comment for the PR, I made a note about resolving the `too-many` warnings if we wanted to. Just didn't want to do a major code change in this PR. 

Sure, I'll go ahead and add `protected-access` to the .plyintrc.

src/ga4gh/vrs/extras/translator.py
jsstevenson: honestly I think also it looks better with "pass" but having to annotate with a pylint ignore is just a nuisance/makes it uglier again. I dunno.

korikuzma: Haha I agree about it looking nicer with pass. What way do you want to go? 😂

korikuzma:
per @ahwagner :
Change 
```from ga4gh.vrs import models, normalize```
to 
```from ga4gh.vrs import models, normalize as do_normalize```

**Comment by korikuzma** [comment]
@ahwagner made your suggested change of renaming the import name. Let me know if there's anything else that needs to be changed

**Review by jsstevenson — APPROVED** [review]

File: src/ga4gh/vrs/extras/localizer.py [hunk 4]
```diff
@@ -45,27 +41,24 @@ def __init__(self):
         self._ana_maps = {k: make_name_ac_map(k) for k in assy_name_to_map_name}
 
 
-    def localize_allele(self, allele):
-        # copy input variant and replace location
-        # N.B. deepcopy leads to recursion errors
+    def localize_allele(self, allele, assembly_name = "GRCh38"):
+        """copy input variant and replace location N.B. deepcopy leads to recursion
```
ahwagner, line 45: The added text looks like a comment more than a docstring

src/ga4gh/vrs/extras/localizer.py
```diff
@@ -45,27 +41,24 @@ def __init__(self):
...
+        errors"""
         allele_sl = ga4gh.vrs.models.Variation(**allele.as_dict())
         del allele_sl._id
-        allele_sl.location = self.localize(allele.location)
+        allele_sl.location = self.localize_named_feature(allele.location, assembly_name)
```
ahwagner, line 49: This doesn't seem right to me by the names and descriptions of these methods. Why would updating a sequence location (parent method) have anything to do with a named feature (child method)? My understanding of named features are annotated regions on a sequence, e.g. gene symbols, protein domains, named exons.

I would need to dig deeper into the code to understand if this is actually correct; and if so we need to reconsider the names and descriptions of these functions.

ahwagner (changes requested):
This looks technically sound to me. I am less certain if these changes are procedurally correct. Specifically, see my comments about the localize mapping to localize_named_feature.

If you are convinced this mapping is correct, then please let @reece and me know so we can do a deeper code review.

src/ga4gh/vrs/extras/localizer.py
korikuzma: From my understanding of this and the [notebook](https://github.com/ga4gh/vrs-python/blob/issue-97/notebooks/Feature%20Based%20Locations.ipynb), these methods converts a Chromsome Location or Gene Location to a Sequence Location

korikuzma: @reece and @ahwagner can you review the changes to the localizer module?

src/ga4gh/vrs/extras/translator.py
```diff
@@ -435,7 +433,7 @@ def ir_stype(a):
 
                 hgvs_exprs += [str(var)]
             except hgvs.exceptions.HGVSDataNotAvailableError:
-                _logger.warning(f"No data found for accession {a}")
+                _logger.warning("No data found for accession %s", a)
```
jsstevenson, line 436: going on the record that pylint is wrong here imo

jsstevenson (approved):

src/ga4gh/vrs/extras/translator.py
jsstevenson: edit: ok, pylint maybe has a point

src/ga4gh/vrs/extras/vcf_annotation.py
```diff
@@ -90,7 +88,7 @@ class SeqRepoProxyType(str, Enum):
     help="The assembly that the `vcf_in` data uses.",
     type=str
 )
-def annotate_click(
+def annotate_click(  # pylint: disable=too-many-arguments
```
theferrit32, line 91: I think if it's fewer than this many, it's okay
https://github.com/chrchang/plink-ng/blob/master/1.9/plink.c#L322

korikuzma: 😂

src/ga4gh/vrs/__init__.py
```diff
@@ -23,4 +23,4 @@
 
 logging.basicConfig(
```
reece, line 24: Libraries should not initialize logging. This should be left to applications. Unexpected or inconsistent behavior is likely with this setup. If I left this here, it must've been during debugging. Up to you whether you want to address it for this PR since it's out of scope.

src/ga4gh/vrs/extras/localizer.py
```diff
@@ -45,27 +41,25 @@ def __init__(self):
         self._ana_maps = {k: make_name_ac_map(k) for k in assy_name_to_map_name}
 
 
-    def localize_allele(self, allele):
+    def localize_allele(self, allele, assembly_name = "GRCh38"):
+        """Converts VRS Allele's named features to sequence location"""
         # copy input variant and replace location
-        # N.B. deepcopy leads to recursion errors
+        # N.B. deepcopy leads to recursion_errors
```
reece, line 47: Okay... I'll bite... why the new underscore? `recursion_errors` is to a symbol.

src/ga4gh/vrs/extras/localizer.py
reece: @korikuzma is correct: the "localization" process is intended to convert a named feature to coordinates. I proposed (and I thought that you and Larry agreed) that we needed to be able to project all features onto a common coordinate system. This operation is necessary in order to identify overlapping features, for example.

src/ga4gh/vrs/extras/localizer.py
```diff
@@ -45,27 +41,25 @@ def __init__(self):
...
-
+        """converts named features to sequence locations"""
         assert loc.type._value == "ChromosomeLocation", "Expected a ChromosomeLocation object"
 
         def _get_coords(m, cb):
             """return (start,end) of band `cb` in map `m`"""
             if cb is None:
                 return None
-            return chr_cb_map[cb][0:2]
+            return m[cb][0:2]
```
reece, line 62: Good catch!

reece (approved):
Looks good to me. I didn't actually run it. Please see comments inline.

src/ga4gh/vrs/extras/localizer.py
korikuzma: I'm actually not sure how an underscore got added. I'll remove

src/ga4gh/vrs/__init__.py
korikuzma: I'll create a new issue for this

src/ga4gh/vrs/extras/localizer.py
ahwagner: Ah, I think I misunderstood; I thought we were "localizing" a sequence location here, but understand now the idea is to localize a feature or chromosome location. That seems appropriately named to me. Thanks @korikuzma and @reece.

ahwagner (approved):

PR #155, opened 2023-01-27
feat: include VRS location digest if identify is True

### LINKED ISSUE #154: Include location ID in translate_from responses
This came up in checking on an AnyVar feature (https://github.com/biocommons/anyvar/issues/29), but if it would be broadly useful to have access to location IDs, we can implement the fix here instead of further down in the stack.

korikuzma:
close #154


jsstevenson (approved):
🙌

PR #166, opened 2023-04-12
ci: python-cqa should run on 0.8 branch

korikuzma:
Or we can just change to `on: [push, pull_request]` like we do in our other vicc repos


jsstevenson:
>  Or we can just change to on: [push, pull_request] like we do in our other vicc repos

I think this would be my preference also

jsstevenson (approved):

PR #172, opened 2023-04-16
Update to VRS 1.3.0 release

ahwagner:


korikuzma (changes requested):
Can you update [this](https://github.com/ga4gh/vrs-python/blob/0.8/.gitmodules#L4) line in .gitmodules to point to the 1.3 branch?

PR #173, opened 2023-04-17
cicd: try having release run on ubuntu-latest

korikuzma:
We're having issues where the release job is hanging when pushing a tag. It might be due to using an old version of ubuntu-latest. Not sure if anything else needs to be changed


jsstevenson (commented):
while we're here, we could also bump setup-python to v4 and checkout to v3 (it'll resolve the warnings)

jsstevenson (approved):

PR #174, opened 2023-04-17
feat: Update submodules to incorporate 1.3 CNV changes

korikuzma:
- Updates versions in GH Actions
- Update README table for vrs-python branch tracking
- `license_file` --> `license_files` (see [here](https://setuptools.pypa.io/en/latest/history.html#id510))
- Variation Normalizer + VCF tests have been failing. These will be addressed at a later time when we officially transition variation normalizer to serverless. 
- Not sure what to tag for the version. We've encountered [this](https://github.com/pypa/setuptools_scm/blob/main/src/setuptools_scm/_modify_version.py#L26-L32) error in the past when dev tags do not end in 0. The latest version is `0.9.0.dev0`. We could do `0.9.0-alpha.0`?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Each line is a file pattern followed by one or more owners.
-*  @andreasprlic @ahwagner @larrybabb
+*  @ga4gh/vr-maintainers @ga4gh/vrs-python-maintainers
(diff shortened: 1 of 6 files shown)


korikuzma:
@ahwagner said 0.9.0-alpha.0 is fine

jsstevenson (approved):

Issue #175, opened 2023-04-20
Error when building mybinder 


SalemBajjali:
I was trying to build https://mybinder.org/v2/gh/ga4gh/vrs-python/main, but it doesn't seem to be working.

korikuzma:
@SalemBajjali Fixed in #176 for the main branch. We'll want to also fix in:
- [ ] 0.9 branch
- [ ] 0.8 branch

korikuzma:
Made PRs from cherry-picked commits from @SalemBajjali & @ahwagner . We can close when these are merged

PR #176, opened 2023-04-20
fixing mybinder build error 

SalemBajjali:
Updated requirements.txt and runtime.txt. #175


korikuzma (approved):
Tested on mybinder and it works! Thanks @SalemBajjali

PR #196, opened 2023-06-17
feat: include VRS location digest if identify is True (0.9 branch)

### LINKED ISSUE #154: Include location ID in translate_from responses

korikuzma:
Close #154 . We did this in #155 for main branch, but did not do it for the 0.9 branch.

Variation Normalizer and VCF tests are expected to fail at the moment.


jsstevenson (approved):
👍

Issue #153, opened 2023-01-27
Consider resolving pylint `too-many` warnings


korikuzma:
We have ignored `too-many-return-statements`, `too-many-branches`, `too-many-arguments`, and `too-many-locals`. We should consider changing the default max in `.pylintrc` and/or resolve these errors rather than ignoring them.

PR #210, opened 2023-07-11
cleanup: add pre-commit + resolve pylint warnings

korikuzma:
See #139 and #149

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,16 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
-  - repo: https://github.com/pycqa/pylint
-    rev: v2.15.9
+  - repo: local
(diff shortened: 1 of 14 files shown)


jsstevenson (approved):
Personally I would be 👍 for blanket ignoring "too few public methods" and "too many locals" in pylintrc

korikuzma:
@jsstevenson We have #153 . Did we want to completely ignore or up the max that we'd allow? I'm fine with either

jsstevenson:
@korikuzma in my opinion, I think "too few public methods" just isn't a useful check for this codebase, and personally I'd be happy to blanket-ignore it.

I looked again at the "too many locals" example and I think it's fine to leave it at the default value for now. It'd be nice to go through en masse (maybe at a future hackathon or during a free week) and try to make refactors to really resolve these issues rather than just ignoring them.

jsstevenson (approved):

PR #215, opened 2023-07-12
Add pydantic classes, not yet integrated as ga4gh.vrs.models

theferrit32:


theferrit32:
hmmm maybe I'll just push

PR #216, opened 2023-07-13
Update to VRS 2.0, with pydantic classes, without python jsonschema objects

theferrit32:


src/ga4gh/vrs/_internal/models.py
```diff
@@ -17,36 +17,555 @@
...
+from enum import Enum
+import inspect
 import logging
+import sys
 import os
-
+import typing
 import pkg_resources
 
+from pydantic import BaseModel, Extra, Field, constr
```
korikuzma, line 29: Can you add pydantic~=1.10.11 to setup.cfg?

jsstevenson:
for reference when we cross this bridge: to make custom basic types in v2, you need to do something like this:

```
from typing_extensions import Annotated  # for python < 3.10
from pydantic import Field, BaseModel

SequenceString = Annotated[str, Field(pattern=r'^[A-Z*\-]*$', description="string description here")]

class ParentClass(BaseModel): 
    other_field: str
    literal_sequence: SequenceString
```

theferrit32:
I will also need to change this mess to look for those fields. Or we note these fields on the classes like we do for the digest fields.
https://github.com/ga4gh/vrs-python/blob/2-alpha/src/ga4gh/vrs/_internal/models.py#L81-L131

PR #234, opened 2023-08-24
feat: allele rle normalization + pin pydantic version

### LINKED ISSUE #204: Support RLE normalization
@ahwagner to provide more details

korikuzma:
@theferrit32 @toneillbroad I don't think RLE normalization was fully added, so this PR will close #204 . The normalize vrs test now passes and have different error messages for the translator ones that failed. The translator tests that previously passed, still pass with these changes.

@ahwagner @larrybabb The [2.0-alpha docs](https://github.com/ga4gh/vrs/blob/86cafd4b1942d78ac8c7701de6d1b33d2f8a145e/docs/source/impl-guide/normalization.rst) has Allele with LSE and RSE (which is not in 2.0-alpha). So I assumed that the LSE section is used for both LSE and RLE. In addition to @theferrit32 or @toneillbroad review, I'd like one of you to review (my brain is on low battery). 

Notes:
- Pin pydantic version to working version. v2.2+ had errors when importing models
- Updated `_normalize_allele` to handle LSE and RLE
  - In previous VRS-Python versions, we did not handle normalization for definite ranges, so I assumed the same still applied
  -  There were checks for different state types, but the only Allele states in 2.0-alpha right now are LSE and RLE so I removed the unnecessary checks
  - Removed normalize test that had RSE

### DIFF
--- a/src/ga4gh/vrs/normalize.py
+++ b/src/ga4gh/vrs/normalize.py
@@ -112,40 +112,62 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
         return input_allele
 
     ival = (start.value, end.value)
-
-    # Get alleles (the sequences to be normalized) for _normalize
     if allele.state.sequence:
         alleles = (None, allele.state.sequence.root)
     else:
         alleles = (None, "")
 
-    # If one of Reference Allele Sequence or Alternate Allele Sequence is empty,
-    # store the length of the non-empty sequence: this is the Repeat Subunit Length
-    len_ref_seq = len(ref_seq[ival[0]: ival[1]])
-    len_alt_seq = len(alleles[1])
-    if not len_ref_seq and len_alt_seq:
-        # Insertion
-        repeat_subunit_len = len_alt_seq
-    elif len_ref_seq and not len_alt_seq:
-        # Deletion
-        repeat_subunit_len = len_ref_seq
+    # Trim common flanking sequence from Allele sequences.
+    try:
+        trim_ival, trim_alleles = _normalize(ref_seq, ival, alleles, mode=None, trim=True)
+    except ValueError:
+        # Occurs for ref agree Alleles (when alt = ref)
+        len_ref_seq = len_alt_seq = 0
     else:
-        repeat_subunit_len = len_alt_seq - len_ref_seq
+        trim_ref_seq = ref_seq[trim_ival[0]: trim_ival[1]]
+        trim_alt_seq = trim_alleles[1]
+        len_ref_seq = len(trim_ref_seq)
+        len_alt_seq = len(trim_alt_seq)
+
+    # Compare the two allele sequences
+    if not len_ref_seq and not len_alt_seq:
+        return input_allele
 
     new_allele = pydantic_copy(allele)
-    try:
-        new_ival, new_alleles = _normalize(ref_seq,
-                                           ival,
-                                           alleles=alleles,
-                                           mode=NormalizationMode.EXPAND,
-                                           anchor_length=0)
 
+    if len_ref_seq and len_alt_seq:
+        new_allele.location.start = _get_new_allele_location_pos(
+            trim_ival[0], start.pos_type
+        )
+        new_allele.location.end = _get_new_allele_location_pos(
+            trim_ival[1], end.pos_type
+        )
+        new_allele.state.sequence = models.SequenceString(trim_alleles[1])
+        if sequence_reference:
+            new_allele.location.sequence = sequence_reference
+        return new_allele
+
+    # Determine bounds of ambiguity
+    try:
+        new_ival, new_alleles = _normalize(
+            ref_seq,
+            trim_ival,
+            alleles=trim_alleles,
+            mode=NormalizationMode.EXPAND,
+            anchor_length=0,
+            trim=False  # Don't need to trim, since we already did it above
+        )
+    except ValueError:
+        # Occurs for ref agree Alleles (when alt = ref)
+        pass
+    else:
         new_allele.location.start = _get_new_allele_location_pos(
             new_ival[0], start.pos_type
         )
         new_allele.location.end = _get_new_allele_location_pos(
             new_ival[1], end.pos_type
         )
+
         new_ref_seq = ref_seq[new_ival[0]: new_ival[1]]
 
         if not new_ref_seq:
@@ -156,26 +178,18 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
                 sequence=models.SequenceString(new_alleles[1])
             )
         else:
-            # Otherwise, return a new Allele using a reference length expression, using
-            # a Location specified by the coordinates of the new ival, a length
-            # specified by the length of the alternate allele, and a repeat subunit
-            # length
+            # Otherwise, return a new Allele using a RLE
             sequence = models.SequenceString(new_alleles[1])
             len_sequence = len(sequence.root)
 
             new_allele.state = models.ReferenceLengthExpression(
                 length=len_sequence,
-                repeatSubunitLength=repeat_subunit_len
+                repeatSubunitLength=len_ref_seq or len_alt_seq
             )
 
             if rle_seq_limit and len_sequence < rle_seq_limit:
-                    new_allele.state.sequence = sequence
+                new_allele.state.sequence = sequence
 
-    except ValueError:
-        # Occurs for ref agree Alleles (when alt = ref)
-        pass
-
-    # Convert IRI back to SequenceReference
     if sequence_reference:
         new_allele.location.sequence = sequence_reference
 
--- a/src/ga4gh/vrs/normalize.py
+++ b/src/ga4gh/vrs/normalize.py
@@ -91,7 +91,7 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
         To exclude `sequence` from the response, set to 0.
         For no limit, set to `None`.
 
-    Does not attempt to normalize Allele's with definite ranges. Will return the
+    Does not attempt to normalize Alleles with definite ranges and will instead return the
         `input_allele`
     """
     allele = pydantic_copy(input_allele)
--- a/src/ga4gh/vrs/normalize.py
+++ b/src/ga4gh/vrs/normalize.py
@@ -87,7 +87,9 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
     :param input_allele: Input VRS Allele object
     :param data_proxy: SeqRepo dataproxy
     :param rle_seq_limit: If RLE is set as the new state, set the limit for the length
-        of the `sequence`. To exclude, set to 0.
+        of the `sequence`.
+        To exclude `sequence` from the response, set to 0.
+        For no limit, set to `None`.
 
     Does not attempt to normalize Allele's with definite ranges. Will return the
         `input_allele`
@@ -185,7 +187,7 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
                 repeatSubunitLength=len_ref_seq or len_alt_seq
             )
 
-            if rle_seq_limit and len_sequence < rle_seq_limit:
+            if (rle_seq_limit and len_sequence < rle_seq_limit) or (rle_seq_limit is None):
                 new_allele.state.sequence = sequence
 
     if sequence_reference:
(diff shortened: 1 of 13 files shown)


src/ga4gh/vrs/normalize.py
```diff
@@ -3,61 +3,170 @@
...
+        )
+        new_allele.location.end = _get_new_allele_location_pos(
+            new_ival[1], end.pos_type
+        )
+        new_ref_seq = ref_seq[new_ival[0]: new_ival[1]]
+
+        if not new_ref_seq:
+            # If the reference sequence is empty this is an unambiguous insertion.
+            # Return a new Allele with the trimmed alternate sequence as a Literal
+            # Sequence Expression
```
korikuzma, line 152: I'm not sure if `reference sequence` is referring to the original reference sequence or new reference sequence given (possible) new ival

korikuzma:
> @ahwagner @larrybabb The [2.0-alpha docs](https://github.com/ga4gh/vrs/blob/86cafd4b1942d78ac8c7701de6d1b33d2f8a145e/docs/source/impl-guide/normalization.rst) has Allele with LSE and RSE (which is not in 2.0-alpha). So I assumed that the LSE section is used for both LSE and RLE.

@ahwagner @larrybabb I think this will cause issues with [hgvs dup del mode](https://github.com/cancervariants/variation-normalization/issues/476#issuecomment-1688329970) in the variation-normalizer if someone chooses the `literal_seq_expr` mode when they want to express using a LSE. If the current changes to the `_normalize_allele` function is correct, I think we should rethink the hgvs dup del mode in variation-normalizer. A proposal would be to just represent using RLE with `sequence` provided.

jsstevenson:
@korikuzma could you make a new issue about the pydantic v2.2+ errors?

tests/test_vrs_normalize.py
```diff
@@ -35,27 +35,65 @@
...
-        "seq_expr": {
-            "location": seq_loc,
-            "type": "DerivedSequenceExpression",
-            "reverse_complement": False
-        },
-        "count": {
-            "type": "Number",
-            "value": 2
-        }
+        "sequence": "",
```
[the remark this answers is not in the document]
ahwagner: This is acknowledging that the entire concept of an Allele (and not a CopyNumber) with ambiguous endpoints is a little absurd. I wonder if we should even be supporting that now that we do CNC / CNX.

src/ga4gh/vrs/normalize.py
```diff
@@ -3,60 +3,192 @@
...
+        `input_allele`
     """
     allele = pydantic_copy(input_allele)
+
+    # Temporarily convert SequenceReference to IRI because it makes the code simpler.
+    # This will be changed back to SequenceReference at the end of the method
     sequence_reference = None
     if isinstance(allele.location.sequence, models.SequenceReference):
         sequence_reference = allele.location.sequence
         allele.location.sequence = models.IRI(sequence_reference.refgetAccession)
```
ahwagner, line 104: Not following this. An IRI isn't guaranteed to have the digest in it, so we should always assume a `SequenceReference`, or have a mechanism for resolving the IRI to get a `SequenceReference`. I might be missing something here, but why not simply remove this block and then revise line 107 to pull from the `digest` field expected in every object?

```suggestion
```

src/ga4gh/vrs/normalize.py
```diff
@@ -3,60 +3,192 @@
...
 
-    sequence = SequenceProxy(data_proxy, allele.location.sequence.root)
+    # Get reference sequence and interval
+    ref_seq = SequenceProxy(data_proxy, allele.location.sequence.root)
```
ahwagner, line 107:
```suggestion
    ref_seq = SequenceProxy(data_proxy, allele.location.sequence.digest)
```

src/ga4gh/vrs/normalize.py
```diff
@@ -3,60 +3,192 @@
...
+            sequence = models.SequenceString(new_alleles[1])
+            len_sequence = len(sequence.root)
+
+            new_allele.state = models.ReferenceLengthExpression(
+                length=len_sequence,
+                repeatSubunitLength=len_ref_seq or len_alt_seq
+            )
+
+            if (rle_seq_limit and len_sequence < rle_seq_limit) or (rle_seq_limit is None):
+                new_allele.state.sequence = sequence
```
ahwagner, line 191: Very minor change that eliminates string duplication operation when `len_sequence >= rle_seq_limit`
```suggestion
            len_sequence = len(new_alleles[1])

            new_allele.state = models.ReferenceLengthExpression(
                length=len_sequence,
                repeatSubunitLength=len_ref_seq or len_alt_seq
            )

            if (rle_seq_limit and len_sequence <= rle_seq_limit) or (rle_seq_limit is None):
                new_allele.state.sequence = models.SequenceString(new_alleles[1])
```

ahwagner (changes requested):
Outstanding work @korikuzma. I have a few general questions for @larrybabb to address and a few minor suggestions for clarity or streamlining. RLE implementation looks sound overall.

For @larrybabb, one major theme raised in this PR is the question of "how do we handle Allele normalization when the Allele Location is specified by Ranges"? To me, these have always seemed to be a shorthand for "I did a targeted region assay and want to craft general statements about copy number in those regions and the potential broader impact they have". I know we allow people to create Alleles with Range-based Locations anyway, but... why? Kori's work here supports those cases and raises interesting questions, e.g. what do we do with definite range intervals? We don't want that discussion to gate this PR but I have created #237 to discuss.

src/ga4gh/vrs/normalize.py
korikuzma: @ahwagner the `.root` is a Pydantic thing, see more about RootModel's [here](https://docs.pydantic.dev/2.1/usage/models/). If I were to commit this change, we would get an AttributeError

tests/test_vrs_normalize.py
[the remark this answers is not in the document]
korikuzma: It's absurd because I added these tests over 2 years ago (when I knew almost nothing). If @ahwagner and @larrybabb  could provide some good test examples, that would be great 😄

src/ga4gh/vrs/normalize.py
korikuzma: This was initially done by @theferrit32 . @ahwagner is the digest expected in every VRS object? In the models.py, the `digest` is set as an optional field. 

I'm trying to figure out why this was done, but I'm now realizing that this is an issue if `allele.location.sequence` is not defined. A `SequenceLocation.sequence` is listed as an optional field. @ahwagner can you explain why this is not a required field?

ahwagner: >@ahwagner is the digest expected in every VRS object?
Yes, this is something that I think @theferrit32 was working on. It is available to every object, and our digest strategy in VRS 2 will compute this for every object, whether or not it is an identifiable object. I believe we were going to be using this for all objects in VRS-Python.

>A SequenceLocation.sequence is listed as an optional field. @ahwagner can you explain why this is not a required field?
Yes, it is optional to use this attribute in JSON Schema, because when used in an `Allele` that is part of a `Haplotype`, the `SequenceLocation.sequence` can be omitted from VRS messages, as they (by definition) will match the `sequence` of the parent `Haplotype` object. However, it is required from the `ga4ghDigest.keys` for creating computed digests, because it is still a critical component of the value of a SequenceLocation. In those cases, it is expected that the system loading the VRS `Haplotype` object would refer to / copy over the `Haplotype.sequence` for the `Haplotype.member[*].location.sequence` properties.

tests/test_vrs_normalize.py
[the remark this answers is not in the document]
ahwagner: Sorry @korikuzma; I didn't mean to indicate that the test was absurd, the notion that @larrybabb and I have not disallowed this use case is absurd. The tests are a good reflection of what we have asked for.

src/ga4gh/vrs/normalize.py
korikuzma: @ahwagner I think there might be some confusion regarding the `digest` field. At the moment, `digest` is optional and does not get computed when a VRS Object is created. Should this field be populated each time a user creates a VRS Object?

korikuzma: @ahwagner is suggesting to switch this condition to IRI. Some test cases: `#seqrefs/myseq123` and `HTTPS://w3id.org/NM_012345`

theferrit32: We will need to dereference the `#seqrefs/myseq123` outside this function because this function only receives the allele to be normalized, it won't have access to the full original document the allele came from, where `#seqrefs/myseq123` can be resolved from. 

We can create another function like `ga4gh_inline` (or something) that takes a JSON document, finds the GA4GH objects in it, and for any field whose value is an IRI that is a relative pointer, inline the object it points to in that field, if the document contains it.

theferrit32: If we do that the `type` fields would then need to be required on the input JSON/dict. They are not currently required because the type fields are defined as literals and when you construct a particular class with some input it assumes it is that type and fills in the type field.

If we don't want this constraint, we could just traverse the input document and replace *all* field values that look like JSON pointers (not just those in fields that are defined as IRIs in the VRS models) with the objects they refer to. This would also let people use the JSON pointer thing in non-model fields. Like if someone has a statement and a custom field they added that isn't in the model, but they want to refer to the variant in the same document from there using a pointer.

korikuzma: @theferrit32 updated comments + tests with deferenced IRIs. Let me know if I need to make any more changes!

src/ga4gh/vrs/normalize.py
```diff
@@ -3,63 +3,192 @@
...
+    Does not attempt to normalize Alleles with definite ranges and will instead return the
+        `input_allele`
     """
     allele = pydantic_copy(input_allele)
-    sequence_reference = None
+
     if isinstance(allele.location.sequence, models.SequenceReference):
-        sequence_reference = allele.location.sequence
-        allele.location.sequence = models.IRI(sequence_reference.refgetAccession)
+        alias = f"ga4gh:{allele.location.sequence.refgetAccession}"
```
theferrit32, line 104: 👍

toneillbroad (approved):

tests/test_vrs_normalize.py
```diff
@@ -66,10 +123,21 @@ def test_normalize_allele(rest_dataproxy):
     allele2 = normalize(allele1, rest_dataproxy)
     assert allele1 == allele2
 
+    allele1_seq_ref = models.Allele(**allele_dict_sequence_reference)
+    allele2_seq_ref = normalize(allele1_seq_ref, rest_dataproxy)
```
theferrit32, line 127: 👍

theferrit32 (approved):

tests/test_vrs_normalize.py
```diff
@@ -75,7 +75,7 @@
     "type": "Allele",
     "location": {
         "type": "SequenceLocation",
-        "sequence": "refseq:NC_000023.11",
+        "sequence": "ga4gh:SQ.w0WZEvgJF0zf_P4yyTzjjv9oW1z61HHP",
```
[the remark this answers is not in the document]
korikuzma: We should add a regex pattern on the IRI class. I can make a new issue for this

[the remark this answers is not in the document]
korikuzma: > An IRI is a reference to another object. It can be of any form. When we say the sequence slot is dereferenced, it means that instead of an IRI, we have a SequenceReference object. This is true for every property in VRS where we allow for an IRI or object.

Okay, I will update the code + tests to always assume a `SequenceReference`

> We SHOULD NOT assume that an IRI takes a specific form (e.g. a refseq or ga4gh identifier) as we do here.

This was just examples for tests. The `SequenceProxy` class will take the input (regardless of refseq/ga4gh/ensembl etc) to get the corresponding sequence.

ahwagner (approved):
Nice!

tests/test_vrs_normalize.py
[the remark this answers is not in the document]
theferrit32: > When we say the sequence slot is dereferenced, it means that instead of an IRI, we have a SequenceReference object.

@ahwagner thanks for this clarification. The Translator class will need to be updated to work like this (doesn't need to be in this PR). Currently it sets the sequence id (`ga4gh:SQ`, not `ga4gh:SQR` ) as the `location.sequence` value

theferrit32 (approved):

PR #244, opened 2023-09-06
Implementing enref/deref for pydantic 2 models

theferrit32:


theferrit32:
Current implementation here pretty much just does what those enref/deref functions did before. It takes a Model instance with identifiable objects in fields and replaces those identifiable fields with their identifier. Right now it just sticks the identifier in as a basic string, not an IRI. Technically this doesn't validate against the model specification, but the code still works. But maybe we should put the identifier in as an IRI so it's a usable Model instance.

This would be wrapping the `_id` on line 57 of `src/ga4gh/core/_internal/enderef.py` with an `IRI()` constructor before returning the identifier. And doing the inverse (IRI -> str) on line 95 where it retrieves the identifier from the object store.

https://github.com/ga4gh/vrs-python/pull/244/files#diff-d01995e424d56d46c0b4afea97a5f1a38e865df8b55b3f6b926f2f2f82a0d2bbR57

```
tests/test_vrs2.py::test_enref
  /Users/kferrite/dev/vrs-python/venv/lib/python3.11/site-packages/pydantic/main.py:301: UserWarning: Pydantic serializer warnings:
    Expected `Union[definition-ref, SequenceLocation]` but got `str` - serialized value may not be as expected
    return self.__pydantic_serializer__.to_python(
```

src/ga4gh/core/_internal/enderef.py
```diff
@@ -73,15 +82,18 @@ def ga4gh_deref(o, cra_map, object_store):
     def _deref(o):
         """depth-first recursive, in-place deref of object; returns id of object"""
         if o.type not in cra_map:
+            _logger.warn(f"{o.type} not in cra_map {cra_map}")
```
korikuzma, line 85: warn is deprecated, see [here](https://docs.python.org/3/library/logging.html#logging.Logger.warning)
```suggestion
            _logger.warning(f"{o.type} not in cra_map {cra_map}")
```

src/ga4gh/vrs/_internal/models.py
```diff
@@ -92,36 +91,29 @@ def pydantic_class_refatt_map():
...
-        lambda c: ('__root__' in c.__fields__
-                   and overlaps(reffable_classes,
-                                flatten_type(c.__fields__['__root__'].type_))),
-        model_classes
-    ))
+    union_reffable_classes = []
+    for model_class in model_classes:
+        if issubclass(model_class, RootModel):
+            flattened_type_annotation = flatten_type(model_class.model_fields["root"].annotation)
+            print("flattened_type_annotation: " + str(flattened_type_annotation))
```
korikuzma, line 104: Remove print
```suggestion
```

korikuzma (changes requested):
Small changes

toneillbroad (approved):

PR #249, opened 2023-09-07
docs: fix UTA readme formatting and wording

jsstevenson:
Noticed a few minor issues in the UTA setup instructions, most notably that some newline chars were getting added into a command where a literal newline was probably intended.

### DIFF
--- a/docs/setup_help/uta_installation.md
+++ b/docs/setup_help/uta_installation.md
@@ -2,11 +2,8 @@
 
         su wag002adm # replace with your admin account name`
 
-2.  Start PostgreSQL:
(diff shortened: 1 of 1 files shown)


korikuzma (commented):
Would we actually be able to move this to biocommons.uta repo? I made an [issue](https://github.com/biocommons/uta/issues/239) for it a while back ago

korikuzma:
> Might be a good idea? FWIW, these instructions make several assumptions about things like the OS and how PostgreSQL is installed, which we might want to resolve if we are adding it to the main repo

Right so including + referencing [this](https://github.com/ga4gh/vrs-python/blob/main/docs/setup_help/m1_mac_setup.md)

jsstevenson:
Going to close this in favor of a more expansive reorganization of docs in #272

PR #252, opened 2023-09-10
build: unpin pydantic version

### LINKED ISSUE #235: Unpin pydantic v2.1.1
Pinning in #234 . We can stick with v2.1.1 for the sake of trying to get work out by plenary. Once we're in a good spot, we should pin to at least the minor version.

korikuzma:
Close #235 More info: https://docs.pydantic.dev/2.3/errors/usage_errors/#root-model-extra

Glad it was a simple fix

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,7 +54,7 @@ install_requires =
     jsonschema>=4.17.3
     numpy
     pyyaml
-    pydantic == 2.1.1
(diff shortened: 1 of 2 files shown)


theferrit32:
Makes sense, thanks for the fix

theferrit32 (approved):

PR #257, opened 2023-09-11
feat!: update vrs 2.0-alpha models 

korikuzma:
Notes:
* Made use of the classes in core-source/vrs-source so we don't repeat code
* Tried to reorganize some models
* Changed enums keys to uppercase
* For ga4gh_identify/ga4gh_serialize/ga4gh_digest, let me know if there's a cleaner way to handle `SequenceReference`
* Update tests since digests changed

### DIFF
--- a/src/ga4gh/vrs/_internal/models.py
+++ b/src/ga4gh/vrs/_internal/models.py
@@ -17,14 +17,14 @@ V1 pydantic: datamodel-codegen --input submodules/vrs/schema/merged.json --input
 V2 pydantic: datamodel-codegen --input submodules/vrs/schema/merged.json --input-file-type jsonschema --output models.py --output-model-type pydantic_v2.BaseModel --allow-extra-fields
 """
 
-from typing import Any, Dict, List, Literal, Optional, Set, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 from enum import Enum
 import inspect
 import logging
 import sys
-import os
 import typing
-from pydantic import BaseModel, ConfigDict, Field, RootModel, constr
+
+from pydantic import BaseModel, ConfigDict, Field, RootModel, constr, model_validator
 
 from ga4gh.core._internal.pydantic import (
     is_identifiable,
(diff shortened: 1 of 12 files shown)


src/ga4gh/vrs/_internal/models.py
```diff
@@ -329,8 +498,7 @@ class SequenceLocation(BaseModel):
         description='The end coordinate or range of the SequenceLocation. The minimum value of this coordinate or range is 0. MUST represent a coordinate or range greater than the value of `start`.',
     )
 
-    class ga4gh:
-        identifiable = True
+    class ga4gh(_Ga4ghIdentifiableObject.ga4gh):
```
theferrit32, line 501: Originally I put these properties in a nested class in order for them to not be seen as related to the top level pydantic properties but still be declared on and accessible on the class. This change seems fine to share some of the structure of the `.ga4gh` nested classes, but if we can think of a cleaner way that'd also be nice.

src/ga4gh/vrs/_internal/models.py
```diff
@@ -133,120 +133,332 @@ def pydantic_class_refatt_map():
...
         None, description='Any primitive or structured object'
     )
 
 
-class Code(RootModel):
-    root: constr(pattern=r'\S+( \S+)*') = Field(
-        ...,
-        description='Indicates that the value is taken from a set of controlled strings defined elsewhere. Technically, a code is restricted to a string which has at least one character and no leading or  trailing whitespace, and where there is no whitespace other than single spaces in the contents.',
-        example='ENSG00000139618',
+class _Entity(BaseModel):
```
theferrit32, line 214: Is there a difference in meaning between classes beginning with an underscore and those not? Some seem kind of like 'internal use' like `_Ga4ghIdentifiableObject`, but others like `DomainEntity` are in the model sources (but not the processed jsonschemas, since they're defined as 'abstract'(?) in the metaschema source files?).

theferrit32 (approved):

src/ga4gh/vrs/_internal/models.py
korikuzma: @theferrit32 Correct, they're meant to be internal. I also realized that maybe I should move the core models to `src.ga4gh.core._internal.models.py`

korikuzma: @theferrit32 pushed change to separate out core models. 

```markdown
  * `from ga4gh.core import core_models`, and refer to models with the
    abbreviated name, e.g., `core_models.Gene` (recommended)

  * `import ga4gh.core`, and refer to models using the fully-qualified
    module name, e.g., `ga4gh.core.core_models.Gene`
```

src/ga4gh/vrs/_internal/models.py
korikuzma: Hmmmm @toneillbroad do you have any suggestions? If not, we can make a separate issue for this and come back to it

toneillbroad: The structures as represented in this PR are good by me.

toneillbroad (approved):

theferrit32 (approved):

PR #264, opened 2023-09-19
fix: identify_all method (rm ga4gh prefixes)

korikuzma:
* ga4gh prefixes were not getting removed during serialization. SequenceReferences are the only types that do not need to be removed

### DIFF
--- a/src/ga4gh/core/_internal/identifiers.py
+++ b/src/ga4gh/core/_internal/identifiers.py
@@ -210,6 +210,10 @@ def identify_all(
     if input_obj is None:
         return None
     output_obj = input_obj
+    if isinstance(input_obj, str):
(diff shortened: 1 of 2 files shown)


larrybabb (approved):

src/ga4gh/core/_internal/identifiers.py
```diff
@@ -210,6 +210,10 @@ def identify_all(
     if input_obj is None:
         return None
     output_obj = input_obj
+    if isinstance(input_obj, str):
+        if input_obj.startswith("ga4gh:") and not input_obj.startswith("ga4gh:SQ"):
```
ahwagner, line 214: I'm not sure I understand this. Why do we not strip the `ga4gh:SQ.` to get the digest from sequence object identifiers?

Issue #275, opened 2023-11-07
Regular expression validation for gnomad format should allow more letters


mingjiecn:
Right now for gnomad_re(https://github.com/ga4gh/vrs-python/blob/593508c6e8229336ca1f53a06f69966020cd68f7/src/ga4gh/vrs/extras/translator.py#L44C3-L44C3), it only allows those Nucleotide symbol:ACGTN. But there are more symbols such as R, Y, K, M, S, W. See all the symbols here: [gnomad_reregular expression validation](https://www.hgmd.cf.ac.uk/docs/nuc_lett.html). Because of the limited symbols, I am not able to generate certain SPDIs from the gnomad format. For example: SPDI `NC_000017.11:83129587:TTGWCACATGATTG:TTG`  from gnomad format `17-83129587-GTTGWCACATGA-G`, and SPDI `NC_000003.12:16894810:W:` from gnomad format `3-16894810-TW-T`

korikuzma:
@mingjiecn I had previously updated the regex in #148 based on the [VCF Specification](https://samtools.github.io/hts-specs/VCFv4.4.pdf)

<img width="830" alt="Screenshot 2023-11-07 at 5 34 35 PM" src="https://github.com/ga4gh/vrs-python/assets/42851808/a7d7cd30-76b8-47c0-ab5b-7a03714a61e1">

I haven't seen any other characters outside of (ACGTN) in gnomad VCF like expressions. @wesleygoar @ga4gh/vr-maintainers  @ga4gh/vrs-python-maintainers Are the other characters that @mingjiecn listed valid for gnomad VCF expressions? If so, I'll go ahead and make this change.

korikuzma:
I do see these characters in the reference sequence. I'll go ahead and make a PR.

korikuzma:
@mingjiecn Created #276 for a fix. Thanks for pointing this out.

mingjiecn:
Thanks for the fast response. Now waiting for the merge and can't wait to use it! @korikuzma

korikuzma:
@mingjiecn Hi, here is the latest release: https://pypi.org/project/ga4gh.vrs/0.8.5/ . Thanks again for making this issue!

PR #276, opened 2023-11-07
fix: gnomad_re should accept all nucleotide characters for ref/alt

korikuzma:
Address #275 for main branch

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -42,7 +42,7 @@ class Translator:
 
     beacon_re = re.compile(r"(?P<chr>[^-]+)\s*:\s*(?P<pos>\d+)\s*(?P<ref>\w+)\s*>\s*(?P<alt>\w+)")
     gnomad_re = re.compile(
-        r"(?P<chr>[^-]+)-(?P<pos>\d+)-(?P<ref>[ACGTURYKMSWBDHVN]+)-(?P<alt>[ACGTURYKMSWBDHVN]+|\*|\.)",
+        r"(?P<chr>[^-]+)-(?P<pos>\d+)-(?P<ref>[ACGTURYKMSWBDHVN]+)-(?P<alt>[ACGTURYKMSWBDHVN]+)",
         re.IGNORECASE
     )
     hgvs_re = re.compile(r"[^:]+:[cgnpr]\.")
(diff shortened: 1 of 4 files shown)


toneillbroad (approved):

ahwagner (approved):
VRS supports all IUPAC codes, this allows all IUPAC codes. Seems reasonable to me!

src/ga4gh/vrs/extras/translator.py
```diff
@@ -41,7 +41,10 @@ class Translator:
     """
 
     beacon_re = re.compile(r"(?P<chr>[^-]+)\s*:\s*(?P<pos>\d+)\s*(?P<ref>\w+)\s*>\s*(?P<alt>\w+)")
-    gnomad_re = re.compile(r"(?P<chr>[^-]+)-(?P<pos>\d+)-(?P<ref>[ACGTN]+)-(?P<alt>[ACGTN]+|\*|\.)", re.IGNORECASE)
+    gnomad_re = re.compile(
+        r"(?P<chr>[^-]+)-(?P<pos>\d+)-(?P<ref>[ACGTURYKMSWBDHVN]+)-(?P<alt>[ACGTURYKMSWBDHVN]+|\*|\.)",
```
ahwagner, line 45: @korikuzma you pointed out that we also use `*` or `.` here. I wouldn't expect to see that in gnomAD, and think we would need to build out advanced methods for handling that in our VCF processor. Are these characters used in gnomAD alt expressions?

```suggestion
        r"(?P<chr>[^-]+)-(?P<pos>\d+)-(?P<ref>[ACGTURYKMSWBDHVN]+)-(?P<alt>[ACGTURYKMSWBDHVN]+)",
```

ahwagner: I know these characters are part of VCF, but that is a file format convention specific to VCF; I do not think this is used by gnomAD, and as far as I can tell isn't required to address #275. So, if we do not have example records from gnomAD where this is needed, I propose we do not add `.` or `*` to the `gnomad_re` method. Our VCF translator _may_ need to handle these at a future date, but until that feature is requested by the community it is not a priority.

korikuzma: @ahwagner works for me! I didn't know if the gnomad_vcf was a combination of gnomad and vcf spec or not.

ahwagner (approved):

src/ga4gh/vrs/extras/translator.py
korikuzma: @ahwagner I'm going to check with the gnomad team on whether `.` exist in their dataset.

korikuzma:
Confirmed with gnomAD team that ALT will never be `.`.

PR #283, opened 2023-11-09
fix: add oneOf property check to Haplotype.members

korikuzma:
https://github.com/ga4gh/vrs/blob/0c9498d4eae1e868ac0b49b7e480fd6a3d67468c/schema/vrs-source.yaml#L169-L177

```yaml
      members:
        type: array
        ordered: false
        minItems: 2
        uniqueItems: true
        items:
          oneOf:
          - $ref: "#/$defs/Allele"
          - $refCurie: gks.core:IRI
```

We were not making the oneOf property check. I believe this is the only place where we had `oneOf` for items but didn't make the check.

### DIFF
--- a/src/ga4gh/vrs/_internal/models.py
+++ b/src/ga4gh/vrs/_internal/models.py
@@ -23,7 +23,7 @@ import inspect
 import sys
 import typing
 
-from pydantic import BaseModel, ConfigDict, Field, RootModel, constr
+from pydantic import BaseModel, ConfigDict, Field, RootModel, constr, field_validator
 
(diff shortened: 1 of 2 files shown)


korikuzma:
I need to check if oneOf property

toneillbroad (approved):

korikuzma:
Closing. This was incorrect on my part.  

https://json-schema.org/understanding-json-schema/reference/array
> An instance is valid against this keyword if is valid against exactly one schema defined by the value of this keyword. The value of this keyword must be an array of valid JSON schemas (objects or booleans).

PR #284, opened 2023-11-10
fix: ensure extra fields present in jsonschema

jsstevenson:
Not really sure why, but when `RootModel` classes are used as attributes in Pydantic models, the `model_json_schema` method seems to get very picky about how things like descriptions and examples are provided. Specifically, it wants them under the `json_schema_extra` param -- otherwise, they won't show up in the JSON schema produced by libraries like FastAPI. This PR makes this change (which coincidentally also silences a sort of annoying warning that starts appearing in more recent Pydantic v2 versions).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,7 +54,7 @@ install_requires =
     jsonschema>=4.17.3
     numpy
     pyyaml
-    pydantic == 2.*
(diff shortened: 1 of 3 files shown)


korikuzma (approved):
Cool, I'll follow this for MetaKB

PR #286, opened 2023-11-18
docs: Update README for running notebooks

manish0kuniyal:
Updated the python version from 3.7 to 3.10.
Provided description about running the notebook on VS Code.

PR #288, opened 2023-11-21
docs: Update README for running notebooks (#286)

korikuzma:
* Updated the python version from 3.7 to 3.10
* Provided description about running the notebook on VS Code
* Included how to run notebooks in Terra with cherry-picked commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Here are some things to try.
 
 Once installed as described above, type
 
-    $ source venv/3.7/bin/activate
+    $ source venv/3.10/bin/activate
     $ jupyter notebook --notebook-dir notebooks/
 
 The following jupyter extensions are recommended but not required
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):

README.md
```diff
@@ -104,7 +104,7 @@ Here are some things to try.
 
 Once installed as described above, type
 
-    $ source venv/3.7/bin/activate
+    $ source venv/3.10/bin/activate
```
theferrit32, line 107: It says "once installed as described above" but this venv and the directory layout `venv/<version>` is created in the makefile. The makefile step `make devready` is below in `Installing for development`. Maybe before the pip install above we should add a step about creating a virtual environment with `python3.10 -m venv venv/3.10`, or add it to this section. 

Should we also put a note that Python 3.10 must be installed? The makefile won't run otherwise.

(I have mixed opinions on tying the makefile to a python version)

korikuzma: @theferrit32 good catch. We should make an issue about reorganizing the readme. I know that people often miss the submodule step during development

I do like the note about making sure python 3.10 is installed

Issue #296, opened 2023-12-12
VRS INFO field definitions output by VCFAnnotator should have correct Number value


ehclark:
The `VCFAnnotator.annotate()` method adds headers to the VCF file that define the INFO fields that are added.  All the definitions currently specify a cardinality of `Number=1`.  But the field value is actually one value for each REF and ALT genotype, which would be `Number=R`.

The resulting VCF file should have the correct cardinality for the INFO fields.  Issue #295 allows the caller to specify whether VRS IDs are computed for REF alleles.  The resulting cardinality for INFO fields should be:
* R - both REF and ALT alleles
* A - ALT alleles only

Issue #295, opened 2023-12-12
Add parameters to VCFAnnotator.annotate() method to limit alleles processed


ehclark:
The current implementation of `VCFAnnotator.annotate()` computes VRS IDs for all REF and ALT alleles for each VCF site.  Add parameters to this method that allow REF or ALT allele VRS computation to be skipped.  This enhancement will allow users to limit VRS ID computation to only those alleles for which it is necessary.

The new parameters added will be optional and default to `True` for backwards compatibility.  The new method signature will be:
```python
def annotate(
    self, vcf_in: str, vcf_out: Optional[str] = None,
    vrs_pickle_out: Optional[str] = None, vrs_attributes: bool = False,
    assembly: str = "GRCh38", compute_for_ref: bool = True, compute_for_alts: bool = True
) -> None:
```

ehclark:
After implementing and thinking through the changes for #296, I have decided to drop the `compute_for_alts` parameter for the following reasons:
* It necessitates using `Number=1` for the VRS INFO fields, which is ambiguous
* There is no real world use case for `compute_for_alts=False`, but it could easily be engineered by preprocessing the input VCF to not include ALT alleles

PR #299, opened 2023-12-14
VCFAnnotator updates to skip REF alleles and emit correct INFO field definitions

ehclark:
Implemented changes for #295 and #296.

The VCFAnnotator tool now accepts an additional parameter that controls whether VRS computation is performed for REF alleles in addition to ALT alleles.  For large scale VRS computation on VCF input, this will allow users to reduce compute time and costs if REF allele VRS IDs are not required.

The VRS INFO field definitions in the output VCF have also been modified to have the correct value for the `Number` attribute: either `R` or `A` depending on whether REF alleles are included.  This change also required passing the VRS INFO field values to the VCF output writer as arrays, rather than strings.


korikuzma (approved):
Looks good to me

jsstevenson (approved):

Issue #297, opened 2023-12-12
VCFAnnotator adds an error message in an INFO field for sites where VRS ID computation fails


ehclark:
When VRS computation for a site fails in `VCFAnnotator.annotate()`, the method raises an exception.  It would be preferable for the error to be caught and an error INFO field added to the VCF site.  This will allow the caller to determine how to handle failures and retain the VRS IDs that can be successfully computed.

korikuzma:
@ehclark can you explain this a little further? Currently, the annotator logs errors when `_from_gnomad` raises an exception.

ehclark:
@korikuzma I want to accomplish 2 things:

1. When VRS ID computation fails for a VCF site, an error message is placed in the output VCF so that the consumer knows affirmatively that an error occurred.  That will allow us to build triage/monitoring features into the consumer system.
2. If an unexpected error occurs processing a VCF site, VRS ID computation for other sites is able to proceed and be returned.  VCF files often have unexpected content that can break code and we will be processing VCF files that originate from many sources.

The main change will be to enclose the VCF site processing code in a try/except:
```python
try:
    vrs_field_data = self._get_vrs_data(
        record, vrs_data, assembly, additional_info_fields, vrs_attributes,
        output_pickle, output_vcf, compute_for_ref, compute_for_alts
    )
except Exception as ex:
    #  ...add error message to VCF...
```

Also, while it is true that `VCFAnnotator._get_vrs_object()` logs errors, in the AnyVar usage, the `_get_vrs_object()` method is overridden and it suppresses exceptions (and looks like it does not log errors either?).  So I will also change this as well so that exceptions propagate up to the top-level try/except at the VCF site level.

PR #301, opened 2023-12-20
Output VRS_Error INFO field for VCF sites where VRS computation fails

ehclark:
Update the implementation that annotates a VCF with VRS IDs such that when an error occurs, a `VRS_Error` INFO field is added to the site instead of the VRS INFO simply being left blank.

The existing implementation caught and suppressed many exceptions at the translation step.  These exceptions are still caught and logged, but then re-thrown to propagate to the top-level VCF site iterator.

With the new behavior, every VCF site will be modified according to the following:
- If an error occurs processing the site, a `VRS_Error` INFO field will be added with the relevant error message
- Otherwise a `VRS_Allele_IDs` INFO field will be added with the VRS ID for the alleles supported by the available translators and blank for those alleles not supported

Percent, newline, tab, semicolon and comma characters in the error messages are escaped with URL (percent) encoding according to the VCF spec.

The unit tests for VCF annotation have been split out into individual tests to make modifications to the VCR cassette fixtures easier to make and debug.  Additional VCR configuration and assertions have been added to enforce a requirement that the cassettes must match exactly what is expected. 

See also #297


larrybabb:
@korikuzma let me know if you have time to review and merge or if I should go ahead and do it.

korikuzma:
@larrybabb I'll take a look now.

src/ga4gh/vrs/extras/vcf_annotation.py
```diff
@@ -134,6 +134,19 @@ class VCFAnnotator:  # pylint: disable=too-few-public-methods
...
+    VCF_ESCAPE_MAP = {
+        "%": "%25",
+        ":": "%3A",
+        ";": "%3B",
+        "=": "%3D",
+        ",": "%2C",
+        "\r": "%0D",
+        "\n": "%0A",
+        "\t": "%09" 
+    }
```
korikuzma, line 149: `:`, `=`, and `\r` are not used. Can we remove them? If so, then instead of

```python
                for search in [ "%", ",", ";", "\t", "\n" ]:
                     err_msg = err_msg.replace(search, self.VCF_ESCAPE_MAP[search])
```

we can do

```python
    VCF_ESCAPE_MAP = {
        "%": "%25",
        ";": "%3B",
        ",": "%2C",
        "\n": "%0A",
        "\t": "%09" 
    }
    VCF_ESCAPE_MAP_KEYS = VCF_ESCAPE_MAP.keys()
...
                for search in VCF_ESCAPE_MAP_KEYS:
                     err_msg = err_msg.replace(search, self.VCF_ESCAPE_MAP[search])
```

korikuzma (approved):
Everything looks good to me. I made a small comment that can be accepted or rejected

src/ga4gh/vrs/extras/vcf_annotation.py
larrybabb: @ehclark do you agree with this feedback? If so, please submit the change if not, please explain why and I'll do the merge.

korikuzma:
@larrybabb You're good to merge

Issue #305, opened 2023-12-26
Improve VRS ID computation performance


ehclark:
This issue is intended to be the beginning of a discussion/analysis of how we might improve the VRS ID computation performance.

### Generating a Baseline
To generate baseline performance/profile data, I ran AnyVar locally and used the annotate VCF REST endpoint to process a single VCF file with 13,275 ref and alt alleles.  AnyVar used a local Postgres storage backend and a local Seqrepo instance.

AnyVar code was the main branch, commit https://github.com/biocommons/anyvar/commit/5c1f37d3f7888ffb0a5d798a81574d7a12deffd4
vrs-python code was the 2-alpha branch, commit 1a2509529becbc836c28015930dd60acc349b1df

My local workstation is a 2018 MacBook Pro with 2.6GHz i7, 32GB RAM and an SSD.  To run AnyVar and generate profiling data:
```bash
cd anyvar
source venv/3.11/bin/activate
SEQREPO_DATAPROXY_URI=seqrepo+file:///usr/local/share/seqrepo/2021-01-29
python3 -m cProfile -o cprofile_baseline.stats -m uvicorn anyvar.restapi.main:app
```

I then used https://github.com/jrfonseca/gprof2dot to convert the profile into an [SVG image](https://github.com/ga4gh/vrs-python/assets/5367470/6817241a-8508-43a2-9258-0f1919695213).

The total time to process the file was 38.7 seconds (excluding server/module startup overhead and request/response I/O).

### Analyzing the Baseline
There are two main branches in the code: generating the VRS object and then serializing the VRS object.  Generating the VRS object is approx 75% of the total processing time; serializing the remaining 25%.  

The VRS object generation appears to be pretty efficient, with the number of method calls closely correlated to the total number of alleles processed and a significant portion of the time consumed by low level file I/O operations in pysam.

Both VRS generation and serialization call into the `ga4gh_identify` method, which is called 66,375 times, or 5 times for each allele.  The methods `ga4gh_digest` and `ga4gh_serialize`, are called 146,025 times.  This indicates the methods are likely being called more times than necessary.

### Improvements
I focused on simple (or perhaps naive?) changes to eliminate serialization where it appeared to be unnecessary.  I identified two places where serialization is done when the identifier generated by the serialization is already available.  Those changes are present in the associated `performance` branch.

[...]

PR #310, opened 2024-01-04
Enhancements to improve VRS computation performance

ehclark:
See also #305

The focus of these changes is to avoid unnecessary VRS ID computation.  There are places in the code where VRS ID computation is done even when a valid VRS ID is already present:
- In `vrs_enref` and `ga4gh_enref` allow the caller to have the VRS ID returned along with the encoded object data
- Provide mechanism for a caller to have `ga4gh_identify` skip VRS ID computation if an ID already exists

The second change is intended for use in situations where the data passed to `ga4gh_identify` originates from a trusted source, such as `VCFAnnotator.annotate()` which is now decorated to skip VRS ID computation.

The combination of these changes yields approximately 20% performance improvement for VCF annotation.

### DIFF
--- a/src/ga4gh/core/_internal/identifiers.py
+++ b/src/ga4gh/core/_internal/identifiers.py
@@ -15,8 +15,11 @@ For that reason, they are implemented here in one file.
 
 """
 
+import contextvars
 import logging
 import re
+from contextlib import ContextDecorator
+from enum import IntEnum
 from typing import Union, Optional
 from pydantic import BaseModel, RootModel
 from canonicaljson import encode_canonical_json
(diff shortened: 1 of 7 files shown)


src/ga4gh/core/_internal/identifiers.py
```diff
@@ -94,10 +144,25 @@ def ga4gh_identify(vro):
...
+        do_compute = False
+        ir = None
+        if when_rule == GA4GHComputeIdentifierWhen.ALWAYS:
+            do_compute = True
+        else:
+            ir = getattr(vro, "id", None)
+            if when_rule == GA4GHComputeIdentifierWhen.MISSING:
+                do_compute = ir is None or ir == ""
+            else:  # INVALID
+                do_compute = ir is None or ir == "" or ga4gh_ir_regexp.match(ir) is None
```
korikuzma, line 157: I think we may be able to do this:

```suggestion
            do_compute = not ir
            if when_rule == GA4GHComputeIdentifierWhen.INVALID:
                do_compute = do_compute or ga4gh_ir_regexp.match(ir) is None
```

korikuzma (approved):

larrybabb (commented):
+1

Issue #312, opened 2024-01-08
Get `0.8` and `2-alpha` up to date with each other


korikuzma:
There are some commits found in `2-alpha` and not `0.8` and vice versa. We should ensure that both are up to date with each other. Additionally, we might want to consider adding documentation to contributing that both branches should be updated with the commit (I typically cherry-pick).

korikuzma:
Closing. I think we are sticking with only bug fixes in 0.8 branch

PR #313, opened 2024-01-08
chore: update test image for uta

korikuzma:
Part of #312

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,12 +17,12 @@ services:
     network_mode: bridge
     ports:
       - 5000:5000
-  
+
   uta:
     # Test:
-    # psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from transcript'
-    # 249909
-    image: biocommons/uta:uta_20180821
+    # psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from uta_20210129b.transcript'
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):
good call

docker-compose.yml
```diff
@@ -17,12 +17,12 @@ services:
...
     ports:
       - 5000:5000
-  
+
   uta:
     # Test:
-    # psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from transcript'
-    # 249909
-    image: biocommons/uta:uta_20180821
+    # psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from uta_20210129b.transcript'
```
larrybabb, line 23: I think the corresponding documentation on the README.md should be changed as well (this took me a minute when I was trying to use the install instructions to figure out.

```
snafu$ psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from transcript'
249909
```

larrybabb (commented):
one README.md change to sync.

docker-compose.yml
larrybabb: i guess I missed the review window ;)

korikuzma: Good catch!!

PR #318, opened 2024-01-09
feat: add intro to terra notebook

korikuzma:


jsstevenson (approved):

korikuzma:
@ga4gh/vrs-python-maintainers I can't access the Terra notebooks and they are probably on VRS 1.x. So I don't know if we want to merge this or not.

theferrit32:
I have never used the Terra notebooks, but it seems like they can configure their own packages by running pip installs from the notebook.

I think it's okay to leave this notebook in but it just might need updates later if someone tries to use it and it doesn't work.

Maybe @larrybabb can confirm whether he can run it in his Terra account

theferrit32 (approved):

PR #328, opened 2024-01-20
fix: use start/end constraints for `digest` and `refgetAccession` fields

### LINKED ISSUE #327: `refgetAccession` property in `SequenceReference` should be exhaustively constrained
In VRS-Python:

```python
refgetAccession: constr(pattern=r'SQ.[0-9A-Za-z_\-]{32}')
```

In VRS:

```yaml
pattern: '^SQ.[0-9A-Za-z_\-]{32}$'
```

Start/end requirements should be added to the regex pattern to ensure conformity.

jsstevenson:
close #327

### DIFF
--- a/tests/test_vrs2.py
+++ b/tests/test_vrs2.py
@@ -1,3 +1,6 @@
+from pydantic import ValidationError
+import pytest
+
 from ga4gh.core import (
     sha512t24u,
     ga4gh_digest,
@@ -36,7 +39,7 @@ allele_383650_dict = {
         "type": "SequenceLocation",
         "sequenceReference": {
             "type": "SequenceReference",
-            "refgetAccession": "ga4gh:SQ.KEO-4XBcm1cxeo_DIQ8_ofqGUkp4iZhI"
+            "refgetAccession": "SQ.KEO-4XBcm1cxeo_DIQ8_ofqGUkp4iZhI"
         },
         "start": 128325834,
         "end": 128325835
@@ -52,7 +55,7 @@ allele_417816_dict = {
         "type": "SequenceLocation",
         "sequenceReference": {
             "type": "SequenceReference",
-            "refgetAccession": "ga4gh:SQ.KEO-4XBcm1cxeo_DIQ8_ofqGUkp4iZhI"
+            "refgetAccession": "SQ.KEO-4XBcm1cxeo_DIQ8_ofqGUkp4iZhI"
         },
         "start": 128325809,
         "end": 128325810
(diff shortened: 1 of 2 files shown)


korikuzma (approved):
Thanks, forgot to do this once [this](https://github.com/ga4gh/vrs/pull/442) VRS PR got merged! I had this change locally, so gene/variation should be good.

korikuzma:
@jsstevenson could we also update the digest pattern in this PR too?

tests/test_vrs2.py
```diff
@@ -166,6 +166,24 @@ def test_vr():
...
+                    "refgetAccession": "ga4gh:SQ.KEO-4XBcm1cxeo_DIQ8_ofqGUkp4iZhI"
+                },
+                "start": 128325834,
+                "end": 128325835
+            },
+            "state": {
+                "type": "LiteralSequenceExpression",
+                "sequence": "T"
+            },
+            "digest": "ga4gh:734G5mtNwe40do8F6GKuqQP4QxyjBqVp"
```
larrybabb, line 185: I thought the plan was for the digest to be only the 'digest' without the "ga4gh:" prefix? @ahwagner  would have the final word on this.

larrybabb (changes requested):
Please validate whether or not the `digest` should include the `ga4gh:` prefix in the value.

tests/test_vrs2.py
jsstevenson: I'll make this section a little clearer

ahwagner: @larrybabb is correct. The `digest` field is for the digest only. The `id` field _may_ contain a ga4gh-prefixed identifier per the VRS 2.0 spec, and VRS-Python _should_ do so.

jsstevenson: not enjoying these mixed signals. (@ahwagner this is in a `with pytest.raises` block)

korikuzma: I think just update the line to: 

```python
"digest": "ga4gh:734G5mtNwe40do8F6GKuqQP4QxyjBqVp"  # this will fail because digest field should NOT include prefix
```

ahwagner: Hey. Sorry, I didn't look at context, only at @larrybabb's comment. I think we're all on the same page: `ga4gh:` in the `digest` field BAD, testing that we catch those errors GOOD.

ahwagner (approved):

larrybabb (approved):
+1

PR #338, opened 2024-02-02
Update serialization code

theferrit32:


jsstevenson (approved):

theferrit32:
rip tests

korikuzma (approved):

ahwagner (approved):

src/ga4gh/vrs/_internal/models.py
```diff
@@ -262,6 +262,13 @@ class ReferenceLengthExpression(_ValueObject):
         None, description='The number of residues in the repeat subunit.'
     )
 
+    class ga4gh(_Ga4ghIdentifiableObject.ga4gh):
```
ahwagner, line 265: Note: this class inheritance (GA4GH Identifiable) is a little confusing, since this object is _not_ GA4GH identifiable but _does_ use the GA4GH keys property to define its serialization. Not sure if it is useful to distinguish this in the VRS-Python class model, so just leaving as a general comment.

theferrit32: Yeah the name may have been decided on before all the use cases were clear. I would be open to renaming `_Ga4ghIdentifiableObject` to `_Ga4ghDigestibleObject` or something else

src/ga4gh/vrs/_internal/models.py
```diff
@@ -271,6 +278,12 @@ class LiteralSequenceExpression(_ValueObject):
     )
     sequence: SequenceString = Field(..., description='the literal sequence')
 
+    class ga4gh(_Ga4ghIdentifiableObject.ga4gh):
```
theferrit32, line 281: Adding inheritance from `_Ga4ghIdentifiableObject.ga4gh` made `LiteralSequenceExpression.ga4gh.identifiable=True` instead of undefined. `LiteralSequenceExpression` should be stripped to these keys but not identifiable (meaning not digested and replaced with its digest during `ga4gh_serialize`).

theferrit32:
I can't remember why we added a `identifiable: bool` property to the ga4gh metadata inner class in models.py, as opposed to just using the existence of a `prefix` value to indicate that a thing was identifiable. Maybe we could go back to the prefix-based way so we don't need to make sure the identifiable property is set to the appropriate value.

theferrit32:
Some tests left failing are due to now omitting `sequence` from the key fields of ReferenceLengthExpression

Just did a quick look at the 3 failures in `tests/extras/test_allele_translator.py` and all 3 are failing for that expected change in the digest.

src/ga4gh/vrs/_internal/models.py
ahwagner: I read through the method and agree this will need refactoring, especially as we have adjusted our serialization strategy to support `ga4gh.keys` in both identifiable and non-identifiable objects.

ahwagner:
I added some code in ... to address https://github.com/ga4gh/vrs-python/pull/338#issuecomment-1924896008. However, I think @larrybabb and I need to write out our strategy for dealing with object digests and serializations now that we have a mixed annotation / value model.

Issue #351, opened 2024-02-22
Repeat subunit length not always correct


ehclark:
It would appear that in the course of completing #342, an important line of code got removed related to correct handling for ambiguous insertions.

The line of code `repeat_subunit_length = math.gcd(len_extended_ref, len_extended_alt)` in `src/ga4gh/vrs/normalize.py` was added in [96d98cd](https://github.com/ga4gh/vrs-python/pull/342/commits/96d98cd34fc1c434d81ae578bbfaaebb41009428), but then dropped in [b51fb6c](https://github.com/ga4gh/vrs-python/pull/342/commits/b51fb6cd3f2c44cbec699dd59f110e081ca2800a).

Without this line, the `repeatSubunitLength` is not always correct.

Example:
```
>>> def get_seq_from_rle(allele, data_proxy):
...     seqId = f"ga4gh:{allele.location.sequenceReference.refgetAccession}"
...     start = allele.location.start
...     end = start + allele.state.repeatSubunitLength
...     subseq = data_proxy.get_sequence(seqId, start, end) # sequence retrieval function, e.g. from SeqRepo
...     c = cycle(subseq)
...     derivedseq = ''
...     for i in range(allele.state.length):
...         derivedseq += next(c)
...     return derivedseq
... 
>>> 
>>> tlr = AlleleTranslator(data_proxy=dp, default_assembly_name="GRCh38", identify=True, normalize=True)
>>> allele = tlr.translate_from('NC_000019.10:g.289464_289465insCACA', 'hgvs')
>>> allele.model_dump(exclude_none=True)
{'id': 'ga4gh:VA.6WEakVj5V1rTYtbNX8Xqx4bMN4pMA6fh', 'type': 'Allele', 'digest': '6WEakVj5V1rTYtbNX8Xqx4bMN4pMA6fh', 'location': {'id': 'ga4gh:SL.L145KFLJeJ334YnOVm59pPlbdqfHhgXZ', 'type': 'SequenceLocation', 'digest': 'L145KFLJeJ334YnOVm59pPlbdqfHhgXZ', 'sequenceReference': {'type': 'SequenceReference', 'refgetAccession': 'SQ.IIB53T8CNeJJdUqzn9V_JnRtQadwWCbl'}, 'start': 289464, 'end': 289466}, 'state': {'type': 'ReferenceLengthExpression', 'length': 6, 'sequence': 'CACACA', 'repeatSubunitLength': 4}}
[...]

PR #352, opened 2024-02-22
Restore repeat subunit length re-computation for normalization

ehclark:
See #351 for more details.

Recompute repeat subunit length to the greatest common denominator of the ref and alt sequence lengths when determining whether the allele state can be represented as a `ReferenceLengthExpression`.

Added an additional HGVS expression for the translator tests that verifies the fix is operating as expected.

### DIFF
--- a/src/ga4gh/vrs/normalize.py
+++ b/src/ga4gh/vrs/normalize.py
@@ -192,6 +192,7 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
         # an RLE allele.
         len_extended_alt = len(new_alleles[1])
         len_extended_ref = len(extended_ref_seq)
+        repeat_subunit_length = math.gcd(len_extended_ref, len_extended_alt)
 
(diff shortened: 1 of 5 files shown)


larrybabb (approved):
+1 good catch

ahwagner:
Hey. Sorry for the delay in responding, have been in meetings all day.

@ehclark this is a good catch, thank you. I did add, and then remove, the gcd function out of concern for how that works in cases where the insertion happens in regions with non-integer repeats of the repeat subunit.

e.g.:

```
CAGCA      <- reference
CAGCAGCAGCA   <- with insertion
```

In this case, the extended ref length = 5 and the extended alt length = 11, and the gcd = 1; so this would end up (erroneously) as an LSE, when in fact this could be explained by insertion of two repeats of reference subunit of length 3.

Seeing that the reversion left open the door to issues like what you raised in #351, however, I have updated the code to capture all of these cases. I am going to restore this branch and reopen the PR with the additional commit, and an additional test that covers this type of event.

Issue #214, opened 2023-07-12
Pass Translator `default_assembly_name` as a method parameter


korikuzma:
Currently, it's an instance variable. I think it makes more sense to have this be a method parameter. The assembly can be changing for each query (which could be often) used and is why I think we should have it as a method param. @ga4gh/vrs-python-maintainers thoughts?

korikuzma:
I forgot that I had added this in a while ago. We allow for `assembly_name` to be set as a key word argument in the translate_from methods.

PR #361, opened 2024-03-11
feat!: parameterize normalize in translator

### LINKED ISSUE #231: Add parameter to perform regex check in translate_from methods
I'm working on an issue in the variation normalizer, where we're first going to use the regex in the Translator, get the position to possibly liftover, and then use the translate from method to return the VRS representation. It would be nice to be able use the existing `match.groupdict()` rather than having this check done twice

korikuzma:
close #231 

@larrybabb I believe this will help with the variation-normalizer issue. 

* new kwarg (do_normalize) in translate_from methods. Set to determine if normalization should be performed

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -54,14 +54,12 @@ class Translator:
         self,
         data_proxy,
         default_assembly_name="GRCh38",
-        normalize=True,
         identify=True,
         rle_seq_limit: Optional[int] = 50
     ):
         self.default_assembly_name = default_assembly_name
         self.data_proxy = data_proxy
         self.identify = identify
-        self.normalize = normalize
         self.rle_seq_limit = rle_seq_limit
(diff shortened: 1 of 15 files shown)


larrybabb (commented):
@korikuzma please verify my understanding that this change is really making sure that a deverloper can more easily turn off the default 'normalization' processing if desired.  Whereas prior to this PR that would have been less straight forward.

larrybabb (approved):
approving this improvement to calling the `translate_from` specializations with a new `do_normalize` option that is default set to `True`.

korikuzma:
> @korikuzma please verify my understanding that this change is really making sure that a deverloper can more easily turn off the default 'normalization' processing if desired. Whereas prior to this PR that would have been less straight forward.

@larrybabb Correct. This is similar to #214 (default assembly name). It doesn't change the functionality, but makes it easier to switch without having to constantly re-set the instance variable.

ehclark (approved):

PR #375, opened 2024-03-26
refactor: simplify validating reference sequence

### LINKED ISSUE #372: DataProxy.is_valid_ref_seq() return signature simplificiation
I think the return object could be simplified here. If `is_valid` is always false when `err_msg` is anything other than an empty string, and always true if `err_msg` is an empty string, does it add any additional information? Could we rephrase this to simply check for invalidity, returning `None` if there are no invalidations, and a string description if it is invalid? Alternatively -- maybe raise an exception if invalid and do nothing if valid? (there's a similar pattern in the popular `requests` library where the `raise_for_status()` method raises an Exception if an HTTP request returns an error code, but is silent otherwise)

_Originally posted by @jsstevenson in https://github.com/ga4gh/vrs-python/pull/366#discussion_r1536120504_

korikuzma:
close #372 . @jsstevenson thanks for the original issue, I think this helps simplify things!

* _is_valid_ref_seq renamed to _validate_ref_seq . This method now does not return anything.


jsstevenson (approved):

jsstevenson (approved):

src/ga4gh/vrs/dataproxy.py
```diff
@@ -144,32 +148,33 @@ def derive_refget_accession(self, ac: str):
...
-    ) -> Tuple[bool, str]:
-        """Return wether or not the expected reference sequence matches the actual
-        reference sequence
 
+    def validate_ref_seq(
+        self, sequence_id: str, start_pos: int, end_pos: int, ref: str,
+        require_validation: bool = True
+    ) -> None:
+        """Determine wether or not the expected reference sequence matches the actual
+        reference sequence
```
jsstevenson, line 157: Just to make the top-level description a bit clearer, maybe something like

```suggestion
        """Determine whether or not the expected reference sequence matches the actual
        reference sequence. Nothing is returned, but invalid results are logged at level WARN
        by default. If ``require_validation`` is ``True``, then invalid data will cause a 
        ``DataProxyValidationError`` to be raised.
```

should add in a `:raises:` clause too (this will become very repetitive but it'll look nicer if we generate an API reference later)

jsstevenson: also you should manually re-type that, I have no idea how the line breaks will actually look

jsstevenson (approved):

larrybabb (approved):
+1 nice job

Issue #380, opened 2024-04-02
Allow hgvs containing `NT` refseq accessions to be parsed


larrybabb:
https://github.com/ga4gh/vrs-python/blob/efbce313f501e86f9cc42565568d9d08a1cfaa85/src/ga4gh/vrs/utils/hgvs_tools.py#L213

There are a handful of NT_ accessions in the clinvar dataset. There's no good reason to prevent this.

PR #383, opened 2024-04-04
feat: add inclusion of "NT" prefixed refseq accessions to hgvs tools

larrybabb:
fix for #380


larrybabb:
@korikuzma this was a simple change. I had it in my earlier PR and then lost it during some of the final code refactoring. Please review and I'll merge it. We can include it in the next alpha release.

korikuzma (approved):

PR #384, opened 2024-04-04
cicd: fix release when multiple branches track same tag

korikuzma:
Our current approach can break when there are multiple branches that track the same tag:

```shell
$ echo $raw
origin/issue-380 origin/main origin/pypi-release-action origin/release origin/vcf-annotator-readme-update
$ branch=${raw##*/}
$ echo $branch
vcf-annotator-readme-update
```

@theferrit32 thanks for pointing out the multiple branches issue! I think this will fix the release action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,8 +16,8 @@ jobs:
         id: get_branch_name
         run: |
           raw=$(git branch -r --contains ${{ github.ref }})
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):
@korikuzma can you leave a brief (1 sentence apiece) description of the problem and the solution in the commit description when you squash? I think some of the innards of this github action can be a little opaque

PR #387, opened 2024-04-05
feat: cnv translator supports uncertain ranges for g. and m.

### LINKED ISSUE #381: Add hgvs parsing support for `Range` based SequenceLocations
The current [hgvs_tools.py](https://github.com/ga4gh/vrs-python/blob/efbce313f501e86f9cc42565568d9d08a1cfaa85/src/ga4gh/vrs/utils/hgvs_tools.py#L16) does not consider the handling of hgvs expressions that have start or end `Range` values. I believe the biocommons/hgvs module support this but we are currently presuming the start/end is always an `int`.

korikuzma:
close #381 

Since we are waiting on HGVS PR to be merged, I used `main-draft` as the base branch.

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -505,16 +505,16 @@ class CnvTranslator(_Translator):
             return position - 1 if do_subtract_1 else position
 
         if isinstance(pos, (SimplePosition, AAPosition, BaseOffsetInterval)):
-            vrs_loc_val = _get_pos_value(pos.base, is_start)
+            vrs_loc_pos_val = _get_pos_value(pos.base, is_start)
         elif isinstance(pos, Interval):
             start_val = _get_pos_value(pos.start.base, is_start and pos.start.base is not None)
             end_val = _get_pos_value(pos.end.base, is_start and pos.end.base is not None)
-            vrs_loc_val = start_val if start_val == end_val else models.Range([start_val, end_val])
+            vrs_loc_pos_val = start_val if start_val == end_val else models.Range([start_val, end_val])
         else:
             err_msg = f"HGVS Location is not supported: {type(pos)}"
             raise ValueError(err_msg)
 
-        return vrs_loc_val
+        return vrs_loc_pos_val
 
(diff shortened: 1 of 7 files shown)


src/ga4gh/vrs/extras/translator.py
```diff
@@ -438,11 +439,14 @@ def _from_hgvs(self, hgvs_dup_del_expr: str, **kwargs):
...
         """
-        # sv = self._get_parsed_hgvs(hgvs_dup_del_expr)
         sv = self.hgvs_tools.parse(hgvs_dup_del_expr)
         if not sv:
             return None
 
+        if sv.type not in {"g", "m"}:
+            err_msg = "Only 'g' and 'm' reference sequences are supported"
+            raise ValueError(err_msg)
+
```
korikuzma, line 449: @larrybabb was this comment for using the CnvTranslator or was it only for range positions?

src/ga4gh/vrs/extras/translator.py
```diff
@@ -10,6 +10,7 @@
 from typing import Optional, Union
 from ga4gh.vrs.dataproxy import create_dataproxy, _DataProxy
 from ga4gh.vrs.extras.decorators import lazy_property
+from hgvs.location import Interval, SimplePosition, AAPosition, BaseOffsetInterval
```
larrybabb, line 13: I think I would prefer that all the code that is related to these hgvs classes be moved to the hgvs-tools.py module. That was my aim with the prior refactoring work (which was half baked but had good intentions).  Could you consider moving this (or posting a new issue for the future to refactor out the hgvs-specific logic from Translator.pyinto hgvs-tools.py?

larrybabb (changes requested):
I'll defer to you. The code looks fine and works the way you have it (afaic).  But I did place a request (which can be deferred to a future issue) if you want.

src/ga4gh/vrs/extras/translator.py
korikuzma: @larrybabb I'll make a separate issue for this

larrybabb (approved):

PR #388, opened 2024-04-08
build: configure build with pyproject.toml

jsstevenson:
I am not sure I can test that this produces a totally valid release, because `setuptools_scm` will [render a version number](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme) that is incompatible with PyPI if you do so from a commit that doesn't have a version tag on it.  I'm pretty sure it's correct, though.


pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+    "pylint",
+    "yapf",
+    "smart_open",
+    # docs
+    "sphinx",
+    "sphinx_rtd_theme",
+    "restview",
+    # idk?  TODO figure it out
+    "ipython",
+    "jedi<0.18",
```
jsstevenson, line 75: truly not sure what these are doing here

setup.cfg
```diff
@@ -1,127 +0,0 @@
...
-max-complexity = 10
-ignore = E129,E221,E241,E251,E303,W291
-
-[tool:pytest]
-# TODO: add src to testpaths to trigger doctests. However, this
-# requires injecting fixtures, which requires conftest.py in src/,
-# which requires packaging changes. Deferred.
-testpaths = src tests
-doctest_optionflags = ALLOW_UNICODE ALLOW_BYTES ELLIPSIS IGNORE_EXCEPTION_DETAIL NORMALIZE_WHITESPACE
-addopts = --cov-config=setup.cfg --cov-report=term-missing --cov=ga4gh
```
jsstevenson, line 127: `setup.cfg` is given as the pytest-cov config target but no actual configuration options are set. I removed it.

setup.cfg
```diff
@@ -1,127 +0,0 @@
...
-
-[build_sphinx]
-all_files  = 1
-
-# http://pep8.readthedocs.org/en/latest/intro.html#error-codes
-[flake8]
-max-line-length = 120
-exclude = tests/*
-max-complexity = 10
-ignore = E129,E221,E241,E251,E303,W291
```
jsstevenson, line 119: Didn't add as a separate `.flake8` file because flake8 isn't ever installed as part of dev dependencies

setup.cfg
```diff
@@ -1,127 +0,0 @@
...
-    ipython
-    jupyter
-    tabulate
-
-
-[aliases]
-test = pytest
-
-[bdist_wheel]
-universal = 1
```
jsstevenson, line 109: This is for retaining compatibility with Python 2 (https://wheel.readthedocs.io/en/stable/user_guide.html#building-wheels) -- we can remove it

pyproject.toml
```diff
@@ -0,0 +1,109 @@
+[project]
+name = "ga4gh.vrs"
+authors = [
+    {name = "Larry Babb", email = "lbabb@broadinstitute.org"},
+    {name = "Reece Hart", email = "reecehart@gmail.com"},
+    {name = "Andreas Prlic", email = "andreas.prlic@gmail.com"},
+    {name = "Alex Wagner", email = "alex.wagner@nationwidechildrens.org"},
```
jsstevenson, line 7: previously this was just Reece, but I went with all of the VR heads for the sake of consistency

pyproject.toml
theferrit32: @jsstevenson I would delete these, I don't know what jedi is used for, I don't think it's necessary. `ipython` is only needed for notebooks

pyproject.toml
```diff
@@ -0,0 +1,109 @@
...
+    # docs
+    "sphinx",
+    "sphinx_rtd_theme",
+    "restview",
+    # idk?  TODO figure it out
+    "ipython",
+    "jedi<0.18",
+]
+notebooks = [
+    "ipython",
```
theferrit32, line 78: I would also delete `ipython` from here, it's pulled in by `jupyter` already

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "vcrpy",
+    "pyyaml",
+    # style
+    "pylint",
+    "yapf",
+    # docs
+    "sphinx",
+    "sphinx_rtd_theme",
+    "restview",
+    "smart_open",  # TODO remove?
```
jsstevenson, line 72: another one -- anybody (@theferrit32, others?) have a clue if this is needed? not finding anything from a quick grep

theferrit32: `smart_open` had been used by Reece in some scripts in some biocommons projects but is not used here, you can remove

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
korikuzma, line 25: Note to update GH actions to also test 3.12

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "pydantic~=2.1",
+]
+
+[project.optional-dependencies]
+extras = [
+    "psycopg2-binary",
+    "biocommons.seqrepo>=0.5.1",
+    "bioutils>=0.5.2",
+    "hgvs>=1.4",
+    "requests",
```
korikuzma, line 54: Should `requests` be in `dependencies` since it's used in dataproxy module (not in extras)?

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "variation"
+]
+requires-python = ">=3.8"
+dynamic = ["version"]
+dependencies = [
+    "bioutils",
+    "canonicaljson",
+    "coloredlogs",
+    "numpy",
+    "pyyaml",
```
korikuzma, line 44: I think we can remove these. I don't see canonicaljson, coloredlogs, or numpy being used anywhere and I think pyyaml belongs in dev and notebooks

```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,105 @@
...
+    "yapf",
+    # docs
+    "sphinx",
+    "sphinx_rtd_theme",
+    "restview",
+    "smart_open",  # TODO remove?
+]
+notebooks = [
+    "jupyter",
+    "tabulate",
```
korikuzma, line 76:
```suggestion
    "tabulate",
    "pyyaml",
```

korikuzma (changes requested):

pyproject.toml
jsstevenson: Yeah looks like a lot of this was removed in 64fee4c1a441e1623cda18734aa28c9137afb14b

jsstevenson: also, numpy????

korikuzma (approved):

PR #394, opened 2024-04-10
refactor: use Annotated with StringConstraints instead of constr

### LINKED ISSUE #350: Replace `constr` with `StringConstraints`/`Field`
Pydantic is deprecating `constr` in favor of `StringConstraints`. It also feels like the `Annotated` strategy of annotating types is preferred, with either a default factory in the Field in the Annotation, or by assigning a default in the class definition.

Can use the `Field`'s automatic handling of StringConstraints fields.

https://docs.pydantic.dev/2.6/api/types/#pydantic.types.constr

I'm struggling to identify any practical difference between these 3 class definitions.

But I think the 2nd option (`MyModel2`, using `Annotated` and `Field` in the type hint) is probably preferred.

```python
from typing import Annotated
import pydantic
from pydantic import BaseModel, Field, StringConstraints

class MyModel1(BaseModel):
    foo: str = Field("MyModel1-foo", description="This is my field", pattern="^[a-zA-Z0-9]+$")
    bar: str = Field(..., description="This is my field", pattern="^[a-zA-Z0-9]+$")

class MyModel2(BaseModel):
    foo: Annotated[
        str,
        Field(..., description="This is my field", pattern="^[a-zA-Z0-9]+$")
    ] = "MyModel2-foo"
    bar: Annotated[
        str,
        Field(..., description="This is my field", pattern="^[a-zA-Z0-9]+$")
    ]

class MyModel3(BaseModel):
    foo: Annotated[
        str,
        StringConstraints(pattern=r"^[a-zA-Z0-9]+$"),
        Field(..., description="This is my field")
    ] = "MyModel3-foo"
    bar: Annotated[
        str,
        StringConstraints(pattern=r"^[a-zA-Z0-9]+$"),
        Field(..., description="This is my field")
    ]


for cls in [MyModel1, MyModel2, MyModel3]:
    m1 = cls(bar="bar")
    assert m1.foo == f"{cls.__name__}-foo"
    assert m1.bar == "bar"
    m2 = cls(foo="foo", bar="bar")
    assert m2.foo == "foo"
    assert m2.bar == "bar"
    try:
        cls(foo="foo-")
        raise AssertionError("Expected a validation error")
    except pydantic.ValidationError as _:
        pass
```

korikuzma:
close #350

* constr is discouraged and will be deprecated in Pydantic 3.0. Per docs, constr returns a type, which doesn't play well with static analysis tools

https://docs.pydantic.dev/2.6/api/types/#pydantic.types.constr--__tabbed_1_1

### DIFF
--- a/src/ga4gh/core/_internal/models.py
+++ b/src/ga4gh/core/_internal/models.py
@@ -10,11 +10,7 @@ Instead, users should use one of the following:
   * `import ga4gh.core`, and refer to models using the fully-qualified
     module name, e.g., `ga4gh.core.core_models.Gene`
 """
-from typing import Any, Dict, List, Literal, Optional, Union
-try:
-    from typing import Annotated  # Python 3.9 and later
-except ImportError:
-    from typing_extensions import Annotated  # For Python 3.8
+from typing import Any, Dict, List, Literal, Optional, Union, Annotated
 from enum import Enum
 
 from pydantic import BaseModel, ConfigDict, Field, RootModel, StringConstraints, model_serializer
--- a/src/ga4gh/core/_internal/models.py
+++ b/src/ga4gh/core/_internal/models.py
@@ -11,9 +11,14 @@ Instead, users should use one of the following:
     module name, e.g., `ga4gh.core.core_models.Gene`
 """
 from typing import Any, Dict, List, Literal, Optional, Union
+try:
+    from typing import Annotated  # Python 3.9 and later
+except ImportError:
+    from typing_extensions import Annotated  # For Python 3.8
 from enum import Enum
 
-from pydantic import BaseModel, ConfigDict, Field, RootModel, constr, model_serializer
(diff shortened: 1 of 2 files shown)


src/ga4gh/core/_internal/models.py
```diff
@@ -11,9 +11,14 @@
     module name, e.g., `ga4gh.core.core_models.Gene`
 """
 from typing import Any, Dict, List, Literal, Optional, Union
+try:
+    from typing import Annotated  # Python 3.9 and later
+except ImportError:
+    from typing_extensions import Annotated  # For Python 3.8
```
korikuzma, line 17: @ga4gh/vrs-python-maintainers are we going to follow biocommons and only support 3 most recent versions of Python?

theferrit32: I would vote for >=3.9

Python 3.8 official EOL is in a few months.

theferrit32 (approved):

PR #396, opened 2024-04-10
fix: get `make test` to run

### LINKED ISSUE #395: `make test` not working
#388 forgot to update makefile commands, so `make test` is not working

```
$ make test
python setup.py pytest
ga4gh/vrs-python/venv/3.10/bin/python: can't open file 'ga4gh/vrs-python/setup.py': [Errno 2] No such file or directory
make: *** [test] Error 2
```

korikuzma:
close #395 

* In #388 we forgot to update the commands to remove setup.py since it was removed. This commit removes references to setup.py and setup.cfg in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,11 +58,6 @@ devready:
 	@echo '#################################################################################'
 
 #=> install: install package
-#=> bdist bdist_egg bdist_wheel build sdist: distribution options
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):
UGH yeah. 

calling pytest with setup.py is discouraged per... a thing in their docs somewhere and should similarly be fixed throughout biocommons

Issue #400, opened 2024-04-17
Passing an invalid variant description to AlleleTranslator.translate_from() throws TypeError


ehclark:
If a variant description that does not match any of the supported formats is passed in to `AlleleTranslator.translate_from()`, a `TypeError` is thrown.
```
>>> tlr.translate_from('BRAF amplification')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/gdh/vrs-python/src/ga4gh/vrs/extras/translator.py", line 98, in translate_from
    o = t(var, **kwargs)
  File "/workspaces/gdh/vrs-python/src/ga4gh/vrs/extras/translator.py", line 305, in _from_hgvs
    return self._create_allele(allele_values, **kwargs)
  File "/workspaces/gdh/vrs-python/src/ga4gh/vrs/extras/translator.py", line 167, in _create_allele
    seq_ref = models.SequenceReference(refgetAccession=values["refget_accession"])
TypeError: 'NoneType' object is not subscriptable
```

This is because the `hgvs_tools.extract_allele_values` can return `None` when the description is not HGVS syntax.  And there is no null defense in `AlleleTranslator._create_allele`.

Invoking `AlleleTranslator.translate_from()` with an invalid description should instead raise a `ValueError` with the message: "Unable to parse data as beacon, gnomad, hgvs, spdi, vrs"

PR #401, opened 2024-04-17
Only attempt to create allele if values were extracted

ehclark:
See #400 

Updated `_from_hgvs()` to only call `_create_allele()` is value extraction was successful. Otherwise returns `None` to indicate description could not be parsed as HGVS.


jsstevenson (approved):
👍

larrybabb (approved):

PR #406, opened 2024-04-30
docs: update readme (install seqrepo, psql cmd, + submodules)

korikuzma:
I haven't tried installing UTA locally, but this is what I noticed when cloning the repo, installing a local seqrepo, + testing out psql command

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,10 @@ First, you must install a local [SeqRepo](https://github.com/biocommons/biocommo
 ```shell
 pip install seqrepo
 export SEQREPO_VERSION=2021-01-29  # or newer if available -- check `seqrepo list-remote-instances`
-sudo mkdir /usr/local/share/seqrepo
+sudo mkdir -p /usr/local/share/seqrepo
 sudo chown $USER /usr/local/share/seqrepo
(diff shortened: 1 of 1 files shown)


theferrit32 (approved):

README.md
```diff
@@ -138,8 +139,8 @@ nearly instantaneous.
 You can test UTA and seqrepo installations like so:
 
 ```shell
-$ psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from transcript'
-249909
+$ psql -XAt postgres://anonymous@localhost/uta -c 'select count(*) from uta_20210129b.transcript'
```
theferrit32, line 142: Does this connection string need to include the password added here?

https://github.com/ga4gh/vrs-python/pull/405/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R27

korikuzma: @theferrit32 I was able to run the command without the password included

korikuzma: @theferrit32 I _think_ it's because anonymous user doesn't have a password and that's why it's not needed

theferrit32: @korikuzma got it, just tried it on my laptop and it also worked

Issue #116, opened 2022-10-24
Update `ga4gh_identify` to take the `ordered` property into consideration


korikuzma:
@ahwagner added an `ordered` property in the VRS schemas for array fields. @ahwagner can you comment more on this field?

ahwagner:
Yes: the `ordered` property of an `array` indicates whether order is meaningful in an array (e.g. in a `ComposedSequenceExpression`) or not meaningful (e.g. in a `Genotype`). If `ordered == True`, then the array elements MUST NOT be sorted during serialization; conversely if `ordered == False`, then the array elements MUST be sorted during serialization.

korikuzma:
Re-opening for 2-alpha

PR #417, opened 2024-07-05
feat!: update vrs/common models

### LINKED ISSUE #409: Implement CisPhasedBlock
And delete Haplotype.

https://github.com/ga4gh/vrs/blob/2.x/schema/vrs/vrs-source.yaml#L148

korikuzma:
close #409 , #116

* updates vrs/gks-common submodules
  * Pydantic models updated to reflect these changes
*  Removed `use_enum_values` from Pydantic config. We didn't use this in all models, so not sure if there was a purpose for it.
* Removed duplicate vrs test

Could definitely get some other eyes on this to make sure I didn't miss anything


src/ga4gh/core/_internal/models.py
korikuzma, line 1: Wondering if we should split these out into core-im vs domain entity modules?

tests/test_vrs2.py
korikuzma, line 1: I deleted test_vrs.py and renamed test_vrs2.py to test_vrs.py. These were essentially the same, but test_vrs2.py had more tests.

korikuzma:
Going to do one more look over in the morning. Then I think I'm ready to mark ready for review.

src/ga4gh/core/_internal/models.py
korikuzma: If so, might do this in a separate PR

jsstevenson (approved):
looks good to me*

*someone who is not particularly involved in any of the changes to these models

PR #421, opened 2024-07-10
refactor: remove _internal directories

### LINKED ISSUE #407: Do away with "private" modules
Models currently live within `_internal` modules, which sort of designates them as quote-unquote "private" unless they're imported into parent modules via `__init__.py`. I'm not sure this is a necessary organizational method, and it might be cleaner to do away with this separation.

korikuzma:
close #407

* Remove _internal directories and move files up a directory


jsstevenson (approved):
👍

PR #422, opened 2024-07-10
refactor!: split gks-common module into domain + entity

korikuzma:
* ga4gh.core.models split into ga4gh.core.domain_models and ga4gh.core.entity_models

Not sure if we want to do this or not, but went ahead and made the PR. If we don't like this, we can close.


jsstevenson (approved):
👍 seems fine

PR #427, opened 2024-07-15
feat: VRS 1.3 digests may be computed for Alleles and Sequence Locations

### LINKED ISSUE #382: Transform variations to previous VRS version
We (me, @jarbesfeld) have been refactoring a mapping pipeline, which previously created VRS 1.3 alleles and haplotypes, to use more up-to-date tooling. This means we're programmatically tied to VRS 2a, but we still need to create VRS 1.3 objects eventually. What we'd like to do is write something like a `.to_version(1.3)` method that produces an equivalent instance of a variation to comply with an older schema (where such equivalences are possible).

We are operating on a pretty quick turnaround time to hit a paper submission, so we'd like to write up what we need soon, but there is a lot of room for bigger conversations about how to expand tooling like this in the future.

ahwagner:
Note: tests only pass for Python 3.12 due to changes in f-string implementation introduced in 3.12. These should be addressed before merge.

closes #382

### DIFF
--- a/src/ga4gh/core/identifiers.py
+++ b/src/ga4gh/core/identifiers.py
@@ -122,7 +122,7 @@ def parse_ga4gh_identifier(ir):
         raise ValueError(ir) from e
 
 
-def ga4gh_identify(vro, in_place='default'):
+def ga4gh_identify(vro, in_place='default', as_version=None):
     """
     Return the GA4GH digest-based id for the object, as a CURIE
     (string).  Returns None if object is not identifiable.
@@ -158,14 +158,14 @@ def ga4gh_identify(vro, in_place='default'):
                 do_compute = not vro.has_valid_ga4gh_id()
 
         if do_compute:
-            obj_id = vro.get_or_create_ga4gh_identifier(in_place)
+            obj_id = vro.get_or_create_ga4gh_identifier(in_place, as_version=as_version)
 
         return obj_id
 
     return None
 
 
-def ga4gh_digest(vro: BaseModel, overwrite=False):
+def ga4gh_digest(vro: BaseModel, overwrite=False, as_version=None):
     """
     Return the GA4GH digest for the object.
 
@@ -181,7 +181,10 @@ def ga4gh_digest(vro: BaseModel, overwrite=False):
 
     """
     if vro.is_ga4gh_identifiable():  # Only GA4GH identifiable objects are GA4GH digestible
-        return vro.get_or_create_digest(overwrite)
+        if as_version is None:
+            return vro.get_or_create_digest(overwrite)
+        else:
+            return vro.compute_digest(as_version=as_version)
     else:
         return None
 
@@ -210,9 +213,12 @@ def collapse_identifiable_values(obj: dict) -> dict:
     return obj
 
 
-def ga4gh_serialize(obj: BaseModel) -> Optional[bytes]:
+def ga4gh_serialize(obj: BaseModel, as_version=None) -> Optional[bytes]:
     """
     TODO find a way to output identify_all without the 'digest' fields on subobjects,
     without traversing the whole tree again in collapse_identifiable_values.
     """
-    return obj.model_dump_json().encode("utf-8")
+    if as_version is None:
+        return obj.model_dump_json().encode("utf-8")
+    else:
+        return obj.ga4gh_serialize_as_version(as_version)
--- a/src/ga4gh/core/identifiers.py
+++ b/src/ga4gh/core/identifiers.py
@@ -137,6 +137,10 @@ def ga4gh_identify(vro, in_place='default', as_version=None):
     - 'never': the vro.id field will not be edited in-place,
         even when empty
 
+    If 'as_version' is set to a version string, other parameters are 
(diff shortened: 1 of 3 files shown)


korikuzma (changes requested):
Some docstrings / comments would be nice

ahwagner:
> Some docstrings / comments would be nice

Added in 06bc1b5.

src/ga4gh/vrs/models.py
```diff
@@ -431,6 +456,29 @@ class SequenceLocation(_Ga4ghIdentifiableObject):
...
+                    if value[0] is None:
+                        result = f'{{"comparator":"<=","type":"IndefiniteRange","value":{value[1]}}}'
+                    elif value[1] is None:
+                        result = f'{{"comparator":">=","type":"IndefiniteRange","value":{value[0]}}}'
+                    else:
+                        result = f'{{"max":{value[1]},"min":{value[0]},"type":"DefiniteRange"}}'
+                else:
+                    raise ValueError(f'{value} is not int or list.')
+                out.append(result)
+            return f'{{"interval":{{"end":{out[1]},"start":{out[0]},"type":"SequenceInterval"}},"sequence_id":"{self.sequenceReference.refgetAccession.split('.')[1]}","type":"SequenceLocation"}}'
```
korikuzma, line 478: This assumes that `sequenceReference` is not null *and* that it is type `SequenceReference`. We should handle cases where `sequenceReference` is null and cases where `sequenceReference` is an `IRI`.

src/ga4gh/vrs/models.py
```diff
@@ -502,6 +564,7 @@ def ga4gh_serialize(self) -> Dict:
 
     class ga4gh(_Ga4ghIdentifiableObject.ga4gh):
         prefix = 'CPB'
+        priorPrefix = {'1.3': 'VH'}
```
korikuzma, line 567: We can probably remove this prior versions aren't supported in this PR
```suggestion
```

src/ga4gh/core/identifiers.py
```diff
@@ -122,7 +122,7 @@ def parse_ga4gh_identifier(ir):
         raise ValueError(ir) from e
 
 
-def ga4gh_identify(vro, in_place='default'):
+def ga4gh_identify(vro, in_place='default', as_version=None):
```
korikuzma, line 125: Probably good to use an enum here if we are expected to support all previous versions (1.2, 1.1, 1.0) in future PRs. If not, we could probably just rename `as_version` to `to_1_3_version` or something and have it be a bool that defaults to `False`. Just so we're not hardcoding.

korikuzma: If supporting multiple versions and using enum, it might be good to fail fast in the beginning of the function and raise an exception if version is not supported.

src/ga4gh/vrs/models.py
```diff
@@ -474,8 +523,21 @@ class Allele(_VariationBase):
         ..., description='An expression of the sequence state'
     )
 
+    def ga4gh_serialize_as_version(self, as_version):
+        """This method will return a serialized string following the conventions for
+        Allele serialization as defined in the VRS version specified by 'as_version`."""
+        location_digest = self.location.compute_digest(as_version=as_version)
+        sequence = get_pydantic_root(self.state.sequence)
```
korikuzma, line 530: We may want to check that `state` is a `LiteralSequenceExpression` and raise exception if not. `LengthExpression` does not have a `sequence` field and not sure if we want `ReferenceLengthExpression` to be converted to `LSE` in 1.3

korikuzma (changes requested):
#428 will resolve failing tests in 3.10 + 3.11.

There are some minor things that we can clean up in separate PRs since you want to get this out fast. Below are my thoughts/suggestions that I think should be addressed in this PR.

src/ga4gh/core/identifiers.py
ahwagner: agree–intent is to support multiple versions down the road and we should leave that door open. Also agree with fail fast approach.

src/ga4gh/vrs/models.py
ahwagner: LSE _or_ RLE is okay here; both have a `sequence` field, and for small variants the RLE `sequence` is populated by default. However, I agree that we want to do some processing here as a separate PR; ideally we would implement the RLE to LSE function (it should work [like this](https://github.com/ga4gh/vrs/issues/426#issuecomment-1828387956)) and just apply that in the case this is an RLE without a defined sequence.

src/ga4gh/vrs/models.py
ahwagner: Yes, agreed.

ahwagner:
@korikuzma I agree with your recommended changes and left comments on each. However, for the purpose of the ASHG demo this is good enough and I am going to merge. Would you kindly take on implementing the above as follow-up PRs?

ahwagner:
Medium-to-low priority: I would like to see these improvements implemented by the end of August.

korikuzma:
@ahwagner Made PR for this. #431 . Supporting other VRS versions will be separate issue.

PR #430, opened 2024-07-16
feat!: update/rename classes to refect public API access and ABC

### LINKED ISSUE #429: Update pydantic models to reflect public api access
We have a lot of classes that start with underscore, such as `class _DomainEntity(BaseModel)`. This indicates that it should only be for internal use. However, many of these classes are used in downstream apps. We should change this to `class DomainEntity(ABC, BaseModel)`.

korikuzma:
close #429

* `_DomainEntity` -> `DomainEntity`
* `_Entity` -> `Entity`
* Abstract classes inherit from ABC

### DIFF
--- a/notebooks/getting_started/1_Quick_Start.ipynb
+++ b/notebooks/getting_started/1_Quick_Start.ipynb
@@ -2,8 +2,15 @@
  "cells": [
   {
    "cell_type": "markdown",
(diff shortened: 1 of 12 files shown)


jsstevenson (approved):
oh heck yeah

PR #431, opened 2024-07-16
feat: add enum for `as_version` + add restrictions to `ga4gh_serialize_as_version`

korikuzma:
Building off of #382 and #427

* Add `PrevVrsVersion` enum to store previous versions of VRS that is supported for computing digests/identifiers
* Updates function signatures + docstrings for ga4gh digest/serialize/identifier
* Adds restrictions to `ga4gh_serialize_as_version`
  * For `SequenceLocation`: `sequenceReference` must be provided and must be a valid `SequenceReference` obj
  * For `Allele`: Only `LiteralSequenceExpression` and `ReferenceLengthExpression` are supported and must provide a `sequence` nonnull attribute.

### DIFF
--- a/src/ga4gh/core/identifiers.py
+++ b/src/ga4gh/core/identifiers.py
@@ -19,7 +19,7 @@ import contextvars
 import logging
 import re
 from contextlib import ContextDecorator
-from enum import IntEnum
+from enum import Enum, IntEnum
 from typing import Union, Optional
 from pydantic import BaseModel, RootModel
 
@@ -56,6 +56,20 @@ class VrsObjectIdentifierIs(IntEnum):
     ANY = 0
     GA4GH_INVALID = 1
     MISSING = 2
+
+
+class PrevVrsVersion(str, Enum):
+    """Define previous VRS versions that are supported for computing digests and
+    identifiers based on the current VRS model
+    """
+
+    V1_3 = "1.3"
+
+    @classmethod
+    def validate(cls, version):
+        if version is not None and version not in cls.__members__.values():
+            err_msg = f"Expected `PrevVrsVersion`, but got {version}"
+            raise ValueError(err_msg)
 
 
 ga4gh_compute_identifier_when = contextvars.ContextVar("ga4gh_compute_identifier_when")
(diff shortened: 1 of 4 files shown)


src/ga4gh/core/identifiers.py
```diff
@@ -58,6 +58,20 @@ class VrsObjectIdentifierIs(IntEnum):
...
+    identifiers based on the current VRS model
+    """
+
+    V1_3 = "1.3"
+
+    @classmethod
+    def validate(cls, version):
+        if version is not None and version not in cls.__members__.values():
+            err_msg = f"Expected `PrevVrsVersion`, but got {version}"
+            raise ValueError(err_msg)
```
korikuzma, line 72: Maybe I should move this to a separate function? It feels weird doing it here.

src/ga4gh/vrs/models.py
```diff
@@ -523,21 +539,31 @@ class Allele(_VariationBase):
...
+        """
         location_digest = self.location.compute_digest(as_version=as_version)
+
+        if not isinstance(self.state, (LiteralSequenceExpression, ReferenceLengthExpression)):
+            err_msg = "Only `LiteralSequenceExpression` and `ReferenceLengthExpression` are supported for previous versions of VRS"
+            raise ValueError(err_msg)
+
         sequence = get_pydantic_root(self.state.sequence)
+
         if sequence is None:
```
korikuzma, line 557: @ahwagner you had [made](https://github.com/ga4gh/vrs-python/pull/427#discussion_r1678545420) this comment. 

> LSE or RLE is okay here; both have a sequence field, and for small variants the RLE sequence is populated by default. However, I agree that we want to do some processing here as a separate PR; ideally we would implement the RLE to LSE function (it should work https://github.com/ga4gh/vrs/issues/426#issuecomment-1828387956) and just apply that in the case this is an RLE without a defined sequence.

However, for this if `sequence` is NOT provided then we'd have to make a call to the SeqRepo dataproxy and that felt weird to introduce. So I just decided to require `state.sequence` to be nonnull.

src/ga4gh/core/identifiers.py
korikuzma, line 72: I was going back and forth on moving this to a function rather than classmethod. I'm fine with whichever.

src/ga4gh/vrs/models.py
ahwagner: That's a good call, I agree.

ahwagner (approved):
This looks outstanding, thanks @korikuzma!

PR #434, opened 2024-07-17
fix: ensure Entity objects return type as str

### LINKED ISSUE #433: Object types return ENUM instead of ENUM.value
When running the Jupyter notebooks, I noticed that VRS objects are being returned with what looks like an ENUM entry for the object type field:

image

Instead, we expect these to be specified by enum.value:

image

korikuzma:
close #433

### DIFF
--- a/src/ga4gh/core/entity_models.py
+++ b/src/ga4gh/core/entity_models.py
@@ -13,7 +13,7 @@ Instead, users should use one of the following:
 from typing import Any, Dict, Annotated, Optional, Union, List
 from enum import Enum
 
-from pydantic import BaseModel, Field, RootModel, StringConstraints, model_serializer
+from pydantic import BaseModel, Field, RootModel, StringConstraints, model_serializer, field_validator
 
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):

src/ga4gh/core/domain_models.py
```diff
@@ -11,14 +11,13 @@
...
 from typing import Literal, Union, List
-from enum import Enum
 
 from pydantic import Field, RootModel
 
 from ga4gh.core.entity_models import DomainEntity
 
 
-class CommonDomainType(str, Enum):
+class CommonDomainType:
```
jsstevenson, line 20: 👍

jsstevenson (approved):

korikuzma:
@jsstevenson I reverted back to str, Enum because I think it makes more sense. We do not want people to be able to reassign these values (whether on purpose or accident) in downstream apps. Lmk if you disagree and I can revert last commit to go back to simple class constant.

jsstevenson (approved):

Issue #225, opened 2023-08-02
Create hashable BaseModel subclass in order to enforce uniqueItems with Set annotation


theferrit32:
We just need to define what to use as the hash value that will work across all the models.

https://github.com/pydantic/pydantic/issues/1303#issuecomment-599712964

@ahwagner if are going to go ahead with making every object digestible in the model and serialize algorithm, that could be a good solution to this, we can just use those digests. (with a special case for custom root types with a non-pydantic type as their root, like IRI being a str)

korikuzma:
@theferrit32 I think this is done [here](https://github.com/ga4gh/vrs-python/blob/main/src/ga4gh/vrs/_internal/models.py#L190-L191). If you agree, we can close

PR #435, opened 2024-07-17
fix: revert Pydantic custom serializer

korikuzma:
* The Pydantic custom serializer resulted in `model_dump` to only include minimal fields for ga4gh serialization. This caused issues in downstream apps that leverage FastAPI

### DIFF
--- a/.github/workflows/python-cqa.yml
+++ b/.github/workflows/python-cqa.yml
@@ -18,17 +18,17 @@ jobs:
       SEQREPO_ROOT_DIR: ./tests/data/seqrepo/latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: actions/cache@v2
+    - uses: actions/cache@v4
       with:
         path: ~/.cache/pip
         key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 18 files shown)


korikuzma:
Oh, I forgot to check `__hash__`

ahwagner (approved):
Yeah, I guess... 😞 

It makes sense to not muck about with the "as json" return, though ideally we would be able to leverage the Rust code in pydantic core for serialization, and not need to introduce a new dependency.

That said, usability is secondary to performance, so I agree this is an improvement. 👍

korikuzma:
@ahwagner I updated the __hash__. Lmk if this is what you think it should be?

ahwagner:
@korikuzma what you wrote is fine. `__hash__` is helpful for comparison operations and indexing, but doesn't play a role in any of the GA4GH identifier methods as far as I can remember. That this now reflects the `ga4gh_digest` method takes a little more compute that what might be strictly necessary, but I doubt it will be meaningful in terms of regular usage. I think we can move forward as is.

korikuzma:
@ahwagner I can revert and use a different hash function, I was just basing off of the current implementation in `main`. @theferrit32 originally made the issue in #225 , so do you have any thoughts?

theferrit32:
@korikuzma I think this is enough for #225 as long as the only types that will be in arrays with `uniqueItems: true` in the schema are identifiable ones. Actually at the moment I'm not seeing any `uniqueItems: true` in the schemas. I think previously it was used in the Haplotype and Genotype classes.

korikuzma:
@theferrit32 ok, I will go ahead and merge and close #225

PR #441, opened 2024-08-09
sync with root-inheritance-fix from VRS

larrybabb:
removed reference to VRS `ValueObject` which was removed in VRS branch `root-inheritance-fix` PR under review.

@korikuzma please check my work and convert let me know how to get this completed and released.


src/ga4gh/vrs/models.py
```diff
@@ -188,7 +188,7 @@ def _recurse_ga4gh_serialize(obj):
         return obj
 
 
-class _ValueObject(DomainEntity, ABC):
+class _ValueObject(Entity, ABC):
```
korikuzma, line 191: I think this is fine, but to be consistent with VRS I wonder if we should remove `_ValueObject` and just extend Entity?

larrybabb:
@korikuzma I need some pointers on how to update the pypi stuff (I think) so that the notebooks can be updated to address the fix to the efo: -> EFO: PR that was made to vrs 2.x recently. I'd like to know how to do that if you have the time to show me.

korikuzma:
@larrybabb Since we no longer use python jsonschema objects, we have to manually update the source code in VRS-Python. For the notebooks, there are some hard-coded EFOs which would also need to be updated manually.

korikuzma:
@larrybabb why was EFO case changed to lowercase in this commit? https://github.com/ga4gh/vrs/commit/cafe826954488a314c73c6e145c3c3e230bb6151

korikuzma (approved):
Either update notebooks in this PR or separate

PR #443, opened 2024-08-28
fix: do not allow `SequenceLocation` to have negative `start`/`end` v…

### LINKED ISSUE #442: SequenceLocation allows for negative start/end coordinates
https://github.com/ga4gh/vrs/blob/2.x/schema/vrs/json/SequenceLocation#L77

https://github.com/ga4gh/vrs/blob/2.x/schema/vrs/json/SequenceLocation#L88

We do not do any validation at the moment. We should add validation checks to follow what is stated in the descriptions.

korikuzma:
…alues

close #442

* Add model_validators to `Range` and `SequenceLocation`

### DIFF
--- a/src/ga4gh/vrs/models.py
+++ b/src/ga4gh/vrs/models.py
@@ -27,7 +27,7 @@ from ga4gh.core import (
 from ga4gh.core.pydantic import get_pydantic_root
 
 from canonicaljson import encode_canonical_json
-from pydantic import BaseModel, Field, RootModel, StringConstraints, ConfigDict, model_validator
+from pydantic import BaseModel, Field, RootModel, StringConstraints, ConfigDict, ValidationInfo, field_validator
 
 from ga4gh.core.pydantic import (
     getattr_in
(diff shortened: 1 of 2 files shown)


src/ga4gh/vrs/models.py
```diff
@@ -454,15 +471,55 @@ class SequenceLocation(_Ga4ghIdentifiableObject):
     )
     start: Optional[Union[Range, int]] = Field(
         None,
-        description='The start coordinate or range of the SequenceLocation. The minimum value of this coordinate or range is 0. MUST represent a coordinate or range less than the value of `end`.',
+        description='The start coordinate or range of the SequenceLocation. The minimum value of this coordinate or range is 0. MUST represent a coordinate or range less than or equal to the value of `end`.',
```
ahwagner, line 474: Like that this needs to be >= 0. There should not be a constraint on `start` < `end` for VRS 2.0, but just noted that this wasn't updated in the VRS 2.0 field definitions. Will address presently.

src/ga4gh/vrs/models.py
```diff
@@ -454,15 +471,55 @@ class SequenceLocation(_Ga4ghIdentifiableObject):
...
     )
     end: Optional[Union[Range, int]] = Field(
         None,
-        description='The end coordinate or range of the SequenceLocation. The minimum value of this coordinate or range is 0. MUST represent a coordinate or range greater than the value of `start`.',
+        description='The end coordinate or range of the SequenceLocation. The minimum value of this coordinate or range is 0. MUST represent a coordinate or range greater than or equal to the value of `start`.',
```
ahwagner, line 478: Like that this needs to be >= 0. There should not be a constraint on `start` < `end` for VRS 2.0, but just noted that this wasn't updated in the VRS 2.0 field definitions. Will address presently.

ahwagner (commented):
start < end should not be enforced, unless `SequenceReference.circular = False`. Though it would be nice to raise a warning if end < start and SequenceReference.circular is undefined. A discussion on this topic was first raised in #362, and later moved ga4gh/vrs#477. The VRS documentation doesn't capture this (ope), but it will be necessary to represent variants spanning the origin.

This isn't described in the VRS documentation, opened ga4gh/vrs#537 to address. Just providing a heads-up here as this PR contains some validation code corresponding to this issue.

korikuzma:
@ahwagner can you re-review?

ahwagner (approved):

PR #444, opened 2024-08-30
chore: update VCF annotator type annotations for 3.10-12

jsstevenson:
making a few small housekeeping PRs local to the VCF module because I'm taking a look at it this morning

~...could also expand this one to the rest of the package possibly~ <- no this is actually a nightmare


jsstevenson:
> ...could also expand this one to the rest of the package possibly

I forgot what a nightmare this is

korikuzma (approved):

PR #446, opened 2024-08-30
feat!: enhance annotator CLI options

jsstevenson:
* Make the input VCF path an [argument](https://click.palletsprojects.com/en/8.1.x/arguments/) rather than an [option](https://click.palletsprojects.com/en/8.1.x/options/). I think this better matches the intent of the tool (and it was the only required option, so might as well make it an arg, right?). Left it at `nargs=1` but I think there's a case for making it `nargs=-1`.
* pathlike args should be `click.Path`/`pathlib.Path`. Add additional configurations, like the input file must exist and the output files must be writable. Didn't make changes further up the stack because I wanted to confine this PR to just the Click interface.
* remove some superfluous/inactive configs
* Touch up help text

### DIFF
--- a/docs/extras/vcf_annotator.md
+++ b/docs/extras/vcf_annotator.md
@@ -20,10 +20,10 @@ The tool uses a SeqRepo data proxy. By default, the local instance at `/usr/loca
 Example of how to run:
 
 ```commandline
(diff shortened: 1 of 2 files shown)


jsstevenson:
# TODO -- gotta update docs/extras/vcf_annotator.md

PR #447, opened 2024-08-30
feat!: declare annotator CLI as console entrypoint

jsstevenson:
Declare the annotator `click` function as a console entry point, enabling `vrs-annotate input.vcf --vcf_out=out.vcf` rather than `python3 src/ga4gh/vrs/extras/vcf_annotation.py input.vcf --vcf_out=out.vcf`.

Open to suggestions on the command name. I'd prefer a structure more along the lines of `vrs-annotate vcf input.vcf --vcf_out=out.vcf` (i.e., `vrs-annotate` is the command, `vcf` is a subcommand) if we think there's a chance that other filetypes could be annotated in the future (in that case, we may want to alter the module name as well). Alternatively, though, the subcommand might be superfluous if the file argument has a suffix indicating the filetype.

### DIFF
--- a/docs/extras/vcf_annotator.md
+++ b/docs/extras/vcf_annotator.md
@@ -10,7 +10,7 @@ The examples run from the root of the vrs-python directory and assumes that `inp
 To see the help page:
 
 ```commandline
-python3 -m src.ga4gh.vrs.extras.vcf_annotation --help
+vrs-annotate-vcf --help
 ```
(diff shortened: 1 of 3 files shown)


korikuzma (approved):

jsstevenson:
initial feelings seem to be supportive of `vcf` subcommand to `vrs-annotate`

PR #450, opened 2024-09-06
feat: enable logging in annotator CLI

jsstevenson:


src/ga4gh/vrs/extras/vcf_annotation.py
```diff
@@ -36,10 +37,55 @@ class SeqRepoProxyType(str, Enum):
 @click.group()
 def _cli() -> None:
     """Annotate input files with VRS variation objects."""
+    logging.basicConfig(
+        filename="vrs-annotate.log",
```
korikuzma, line 41: This made me realize that we should probably rename class / module to vrs-annotate if we plan on supporting more than just vcfs. Can create a new issue for this

jsstevenson: @korikuzma yeah I thought about doing that in the previous PR and wasn't sure about it, but would be happy to make a PR for it if we think it makes sense to do so now

PR #453, opened 2024-10-28
feat!: update gks-common / vrs models

### LINKED ISSUE #452: Update core + vrs models
GKS-Common + VRS models have had significant updates. We should update the corresponding core and vrs Pydantic models accordingly.

korikuzma:
close #452


toneillbroad (approved):

korikuzma:
Sorry... now I'm done

tests/validation/test_schemas.py
```diff
@@ -25,7 +25,7 @@ class GKSSchemaMapping(BaseModel):
     base_classes: set = set()
     concrete_classes: set = set()
     primitives: set = set()
-    schema: dict = dict()
+    schema: dict = {}
```
jsstevenson, line 28: as a side note, I tried to properly add ruff to vrs-python a few months ago, but the import sort ended up introducing a circular dependency. If it's not doing that anymore, then we could probably get it added in.

jsstevenson (approved):

PR #454, opened 2024-10-29
cicd: add basic precommit hooks

jsstevenson:
* Add precommit + some minimally controversial hooks. Add a note in README.
* Run hooks over existing codebase. There's a few instances of trailing whitespace and incorrect line endings.
* Run checks for formatting-related hooks in CI


Makefile
```diff
@@ -47,7 +47,8 @@ venv/%:
 #=> develop: install package in develop mode
 .PHONY: develop setup
 develop setup:
-	pip install -e .[dev,extras,notebooks]
+	pip install -e .[dev,extras,notebooks]; \
+	pre-commit install
```
jsstevenson, line 51: this seems to work for (and make sense to) me but I'm always wary of makefile environment construction soooo

submodules/README.md
korikuzma, line 1: We shouldn't run checks in submodules dir

korikuzma (changes requested):
> Not sure if this should be mentioned explicitly in the readme?

This would be nice

.github/workflows/python-cqa.yml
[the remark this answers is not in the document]
jsstevenson: Yes -- was originally using it but it a) doesn't seem particularly necessary, this is like a 4 line action, b) it's maintenance-only, I'm not sure how quick they would be to upgrade dependencies. They suggest `pre-commit.ci lite` but I think this is also some extra heft that we don't necessarily need, I think workflows can be abstracted enough as it is.

submodules/README.md
jsstevenson: pre-commit doesn't run on anything in a submodule, unless I'm mistaken. The only affected file in that directory is `submodules/README.md`.

korikuzma: Ope, my bad. Only saw submodules dir when I was reviewing quickly

korikuzma (approved):

PR #461, opened 2024-11-04
fix: don't write digest property when `in_place=False`

jsstevenson:


jsstevenson:
don't do this
* too many "modes"
* just make a subsequent method to strip digests/etc

PR #467, opened 2024-11-27
feat: make VRS schema release info accessible in VRS-Python

### LINKED ISSUE #85: retrieve schema version
Currently we can use VRS Python to retrieve the software version, but not the version of the VRS schema being used. We should add in a method to determine this.

jsstevenson:
(draft: don't worry about reviewing yet)

close https://github.com/ga4gh/vrs-python/issues/85

* use a hook during setup to grab the most recent release tag from the VRS submodule, then write it as an importable Python module (not necessarily the most elegant way to do this, but I think it's the only realistic option)
* This draft uses the scheme used by [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme) to append commit hash/rev distance when the most recent tag isn't applied to the checked-out commit. I assume that the VRS submodule is tied to a branch and not a tag/commit due to its development status, and maybe this will be unnecessary once that stabilizes.
* very, very low priority. Was just something I was thinking about yesterday.

Current status: for the life of me, I cannot get this to work in CI/CD. Seems to be something about how the GitHub checkout action works and what it does and doesn't get from the `.git` folder. Not sure how to move forward.

### DIFF
--- a/.github/workflows/python-cqa.yml
+++ b/.github/workflows/python-cqa.yml
@@ -21,6 +21,8 @@ jobs:
     - uses: actions/checkout@v4
       with:
         submodules: recursive
+        fetch-tags: true
+        fetch-depth: 1
     - uses: actions/cache@v4
       with:
         path: ~/.cache/pip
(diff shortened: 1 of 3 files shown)


jsstevenson:
going with the slightly less elegant #543 instead

PR #483, opened 2025-01-30
style: lint and format with ruff

### LINKED ISSUE #411: pylint/yapf -> ruff
ruff hegemony 😎

jsstevenson:
close #411 

- [x] lots left to do, some farmed out to new issues
    * I made some decent progress in `vrs/models.py` but it's just an utter minefield so I opted to exclude it from linting entirely. It'd be nice to return at a later point but it's gonna take some real attention, particularly given the way that equivalent type annotations seem to have some runtime effects here.
    * Some other lint groups are blanket ignored for a couple of different files, mostly related to docstrings and type annotations for modules that I'm not particularly familiar with. It'd be good to make progress on them in the future.
    * Made a few specific issues about particular things that will entail minor runtime changes to correctly fix. Otherwise, every change made here should be more or less irrelevant for anything other than style
- [x] add to CICD/precommit
- [x] sort out circular import issue

open questions:

- [x] should the test action only run if lint passes?

### DIFF
--- a/.github/workflows/python-cqa.yml
+++ b/.github/workflows/python-cqa.yml
@@ -2,13 +2,26 @@ name: Python CQA
 
 on: [push, pull_request]
 
-
-# TODO: separate lint and test jobs; lint first, test "needs" lint
-# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs
-
 jobs:
-  test:
 
+  lint:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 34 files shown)


jsstevenson:
> Yes, following https://github.com/GenomicMedLab/software-templates/issues/50

https://www.youtube.com/watch?v=xs04iBJeFEo

korikuzma (approved):
🧼

PR #484, opened 2025-01-31
featremove unsafe uses of assert

### LINKED ISSUE #480: Remove all `assert` instances in library code
The `normalize()` function includes a use of `assert` to check that the incoming value is a Pydantic object. (also found in `dataproxy.py` and `core/enderef.py`)

This is quite bad! `asserts` are ignored when Python is run in "performance" mode -- as it is in a lot of popular web servers. These should be removed. See eg https://docs.astral.sh/ruff/rules/assert/

jsstevenson:
close #480


korikuzma (approved):
Consider updating docstrings to include :raises:

korikuzma (approved):

PR #494, opened 2025-02-05
feat!: break annotator tools into separate submodule

jsstevenson:
* We have long bristled at the naming incongruence between `vcf_annotation.py` and `translator.py`. This makes the former consistent with the latter.
* Anticipating the possibility that we may want to annotate other kinds of files in the future, this makes `annotator` a package within `extras`, containing a `vcf` module. Also breaks the CLI into a separate module.

proposed module structure:

```
src/ga4gh/vrs/extras
├── __init__.py
├── annotator
│   ├── __init__.py
│   ├── cli.py
│   └── vcf.py
├── decorators.py
├── object_store.py
└── translator.py
```

### DIFF
--- a/tests/extras/test_annotate_vcf.py
+++ b/tests/extras/test_annotate_vcf.py
@@ -1,13 +1,14 @@
 """Ensure proper functionality of VCFAnnotator"""
 
 import gzip
+import logging
 import re
 from pathlib import Path
 
 import pytest
 
 from ga4gh.vrs.dataproxy import DataProxyValidationError
-from ga4gh.vrs.extras.vcf_annotation import VCFAnnotator, VCFAnnotatorException
+from ga4gh.vrs.extras.annotator.vcf import VCFAnnotator, VCFAnnotatorException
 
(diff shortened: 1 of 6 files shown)


tests/extras/test_annotate_vcf.py
```diff
@@ -168,6 +169,9 @@ def test_annotate_vcf_input_validation(vcf_annotator):
 @pytest.mark.vcr
 def test_get_vrs_object_invalid_input(vcf_annotator, caplog):
     """Test that _get_vrs_object method works as expected with invalid input"""
+    # some tests are check
```
korikuzma, line 172: Don't think this is complete?

korikuzma (approved):

tests/extras/test_annotate_vcf.py
jsstevenson: Yeah not sure what happened there

jsstevenson: The author was assassinated mid-sentence

korikuzma (approved):

PR #497, opened 2025-02-05
 feat!: move, reorganize field const variables 

jsstevenson:
Another easier one.

* `str.translate` is a more straightforward way to do a multi pass mapping on a string
* move consts outside of the class. It probably doesn't matter but it made more sense to me to put them there in case another module wants to refer to them.

### DIFF
--- a/src/ga4gh/vrs/extras/annotator/vcf.py
+++ b/src/ga4gh/vrs/extras/annotator/vcf.py
@@ -33,28 +33,31 @@ class SeqRepoProxyType(str, Enum):
     REST = "rest"
 
 
+# Field names for VCF
+VRS_ALLELE_IDS_FIELD = "VRS_Allele_IDs"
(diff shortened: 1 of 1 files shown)


src/ga4gh/vrs/extras/annotator/vcf.py
```diff
@@ -33,29 +33,32 @@ class SeqRepoProxyType(str, Enum):
     REST = "rest"
 
 
+# Field names for VCF
```
korikuzma, line 36: Wonder if we should just make VCF Info Enum

korikuzma (approved):

src/ga4gh/vrs/extras/annotator/vcf.py
jsstevenson: @korikuzma good point, I'll do that tomorrow

PR #498, opened 2025-02-05
feat!: simplify dataproxy construction interface

jsstevenson:
Rather than accepting various combos of params at the CLI and the annotator constructor, take a single URI and build the dataproxy from it.

* this matches how things are done elsewhere in Biocommons and the VRS Python translator class
* it's a bit nicer than having multiple arguments that need to be set based on the value of another argument
* it centralizes data proxy construction into one place rather than doing it in a different way

One funny thing -- as a result, some slightly different test configuration is needed. These changes make it so a single dataproxy instance can be used for each test case. However, because the tests all assume the same input VCF, that would mean that every test after the first one would get cache hits for all seqrepo requests. I have a separate PR addressing the bigger way that the cache is influencing cassettes, but as a temporary measure I made a function-scoped dataproxy fixture just for this module.

I do think it's nice to have a default value for CLIs so I kept one. Considered setting a default on the `create_dataproxy` function but it'd be annoying to create further drift between code there and the dataproxy module in SeqRepo.

Also added a "ping" to SeqRepo at the beginning of the annotator. A downside of suppressing and logging *every* error when doing annotation is that you might've provided the wrong SeqRepo REST URI and it'll just dutifully annotate it as an error to every VCF row. If using a file dataproxy the constructor will catch stuff like a nonexistent seqrepo dir early, but otherwise no connection-based things are known upfront for the REST dataproxy. Realistically, we should be doing better than a blanket `except` but that can be a future project.

I will call this the last issue for #482


jsstevenson:
another TODO: dataproxy should ping upon construction

theferrit32 (approved):

pyproject.toml
```diff
@@ -49,6 +49,7 @@ extras = [
     "hgvs>=1.4",
     "dill~=0.3.7",
     "click",
+    "requests",
```
korikuzma, line 52: I don't think this is needed since it's included in required `dependencies` in line 40
```suggestion
```

korikuzma (approved):

pyproject.toml
jsstevenson: Yup, missed that

PR #499, opened 2025-02-06
test: add coverage for dataproxy construction inputs

jsstevenson:
* check invalid inputs
* ensure env var works


tests/extras/test_dataproxy.py
```diff
@@ -20,3 +24,24 @@ def test_data_proxies(dp, request):
         "ga4gh:SQ.v_QTc1p-MUYdgrRv4LMT6ByXIOsdw3C_", start=0, end=50
     )
     assert seq == "CCTCGCCTCCGTTACAACGGCCTACGGTGCTGGAGGATCCTTCTGCGCAC"
+
+
+def test_invalid_data_proxy_uri():
```
korikuzma, line 29: Do we need to test if uri is none?

korikuzma (approved):

PR #502, opened 2025-02-07
feat!: refactor annotator as ABC, add NDJSON annotation + support optional side effects

### LINKED ISSUE #474: Add callback support for VCF annotator
> Second, this may be an opportunity to improve the relationship between AnyVar and vrs-python. The AnyVar VcfRegistrar extends VCFAnnotator and overrides the internal _get_vrs_object() method. The only reason for doing so is to integrate with the AnyVar object store. But then AnyVar does not actually re-implement all the vrs-python functionality, which means there will be continual drift between the two implementations. It seems like whether you annotate a VCF with VRS IDs directly in vrs-python or via AnyVar that the behavior and result should be the same. That would make the tools easier to adopt and use. So perhaps a re-design of the vrs-python VCFAnnotator to provide a listener or callback. For example:

```python
class VCFAnnotator:
  # ...existing implementation...

  def on_vrs_object(self, vcf_coords: str, vrs_object: VrsObject) -> bool:
    """Called for each calculated VRS object prior to it being added to the VCF file.
    :param vcf_coords: The VCF coordinates of the VRS object
    :param vrs_object: The computed VRS object
    :return:  True if the VRS object should be added to the output VCF, False otherwise
    """
    return True

# And in AnyVar
class VcfRegistrar(VCFAnnotator):
  def on_vrs_object(self, vcf_coords: str, vrs_object: VrsObject) -> bool:
    self.av.put_object(vrs_object)
    return True
```

originally posted by @ehclark in https://github.com/ga4gh/vrs-python/issues/470#issuecomment-2531684852

jsstevenson:
close #474 

Refactor the Annotator class to better support configurable outputs and side effects beyond an annotated VCF. Specifically, the class is now an Abstract Base Class, `AbstractVcfAnnotator`, and implementations must define three methods:

* `on_vrs_object`: perform filtering/transformation/side effects on VRS objects that have been translated from VCF coords. For example, attach additional extensions or mappings, or upload to a DB. Called every time the translator produces a successful VRS allele.
  * Eugene had originally proposed this solely for producing side effects (like DB uploads). I thought it might be helpful to a) enable transformations/more annotations (maybe do something with the expression added by the translator, or don't pass along for VCF annotation if it meets certain conditions). I don't have any specific applications in mind, though, so if it doesn't seem well-thought out, that's because it isn't.
* `on_vrs_object_collection`: do something with the aggregation of all VRS alleles collected during VCF annotation, such as dump them to a file. Called once after VCF ingestion is complete, but only if the class variable `collect_alleles` is `True`.
* `raise_for_output_args`: double-check that *some* kind of output has been declared in `annotate()`. This is here because there was a similar check in `annotate()` previously. The idea is to force a fast failure if you aren't going to be producing any kind of output. I don't feel particularly tied to keeping this, though.

The existing pickle file dump is modified slightly to use VRS IDs as keys (unsure why the other thing was being used previously, could change back if necessary). It's refactored to be an optional add-on, and an additional option to output an NDJSON dump is added. A basic implementation incorporating all of this is defined in the class `VcfAnnotator`. The CLI is updated to use this class.

See https://github.com/biocommons/anyvar/blob/b57f400c1fe4b821828847d906ac0a5246308d36/src/anyvar/extras/vcf.py for an external implementation.

Some potential issues:
* In general, reliance on `kwargs` to pass arguments to the child class methods -- it's a little clunky obviously, both with use and documentation.
* since retaining all constructed alleles might be costly for speed and memory, it can be disabled/enabled by an implementation with the class variable "collect_alleles". It's disabled by default, so if you tried to add functionality like dumping to a file and missed that you need to change the class variable, it wouldn't do anything. Ideally there would be some way to raise an abc error of some kind, idk.
* Previously, the VRS allele collection that's retained while ingesting VCFs (it was named `vrs_data` before, I'm trying a name like `allele_collection`?)  was retaining stringified dict dumps of alleles (i.e. `str(allele.model_dump(exclude_none=True))`). I'm not totally sure why this was the case, so I changed it to just hold onto the pydantic objects and defer decisions about serialization etc to `on_vrs_object_collection`. I am not sure if this causes memory issues with extremely large VCFs.

### DIFF
--- a/src/ga4gh/vrs/extras/annotator/vcf.py
+++ b/src/ga4gh/vrs/extras/annotator/vcf.py
@@ -72,7 +72,17 @@ def dump_alleles_to_ndjson(
 
 
 class AbstractVcfAnnotator(abc.ABC):
-    """Abstract class for VCF annotation with VRS."""
+    """Abstract class for VCF annotation with VRS.
+
+    Child classes should implement all abstract methods, though some may be less relevant
+    and can just be stubbed out (for example, if the desired side effect for each VRS
+    object is something like a REST request, there may be no need for extra cleanup in
+    ``on_vrs_object_collection``).
+
+    Critically, implementations that intend to do something with all VRS objects at once
+    following ingestion (e.g. writing a PKL dump) need to set the class variable
+    ``collect_alleles`` to ``True``.
+    """
 
     collect_alleles: bool = False
 
@@ -313,12 +323,12 @@ class AbstractVcfAnnotator(abc.ABC):
                 "Exception encountered during translation of variation: %s", vcf_coords
             )
             raise
-        if vrs_obj is None:
+        if vrs_obj is not None:
+            vrs_obj = self.on_vrs_object(vcf_coords, vrs_obj, **kwargs)
+        else:
             _logger.debug(
                 "None was returned when translating %s from gnomad", vcf_coords
             )
-        else:
-            vrs_obj = self.on_vrs_object(vcf_coords, vrs_obj, **kwargs)
 
         if allele_collection is not None and vrs_obj:
             allele_collection.append(vrs_obj)
--- a/src/ga4gh/vrs/extras/annotator/vcf.py
+++ b/src/ga4gh/vrs/extras/annotator/vcf.py
@@ -1,5 +1,6 @@
 """Annotate VCFs with VRS identifiers and attributes."""
 
+import abc
 import logging
 import pickle
 from enum import Enum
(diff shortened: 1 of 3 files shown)


ehclark (approved):

jsstevenson:
@quinnwai

src/ga4gh/vrs/extras/annotator/vcf.py
```diff
@@ -284,18 +353,19 @@ def _get_vrs_object(
...
         compute_for_ref: bool = True,
         require_validation: bool = True,
+        **kwargs,
     ) -> dict:
         """Get VRS data for record's reference and alt alleles.
 
         :param record: A row in the VCF file
-        :param vrs_data: Dictionary containing the VRS object information for the VCF.
-            Will be mutated if `output_pickle = True`
+        :param allele_collection: Dictionary containing the VRS object information for
```
quinnwai, line 367: should this be List not Dictionary?

PR #506, opened 2025-02-11
fix: forbid extra attributes when additionalProperties is set to False

### LINKED ISSUE #505: Pydantic models do not enforce JSON Schema `additionalProperties: false`
https://json-schema.org/understanding-json-schema/reference/object#additionalproperties

The source.yaml files do not always have this set (default is true in JSON Schema), but the MSP updates the JSON schema files with `additionalproperties: false`. We should forbid extra attributes on our pydantic models (https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra)

korikuzma:
close #505

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "submodules/vrs"]
 	path = submodules/vrs
 	url = https://github.com/ga4gh/vrs.git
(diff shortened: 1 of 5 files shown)


larrybabb (approved):
seems good to me.

PR #508, opened 2025-02-12
fix(ci): simplify main branch check

korikuzma:


jsstevenson (approved):

korikuzma:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#determining-when-to-use-contexts

Found example in docs so it should work?

jsstevenson:
@korikuzma I think that commit should do it (fingers crossed)

PR #510, opened 2025-02-12
fix(ci): revert #508 and #509 + exclude HEAD from git branch output in pypi release job

korikuzma:


korikuzma:
This one should work because I made this change in Cat-VRS Python https://github.com/ga4gh/cat-vrs-python/pull/10

jsstevenson (approved):

PR #539, opened 2025-04-07
cicd: do not install setuptools

jsstevenson:
We were missing an issue upstream about a deprecated stdlib module because `setuptools` comes with an in-place fix -- but `setuptools` wasn't an explicit dependency declared for downstream users. I think we should just start with whatever `setup-python` gives us unless there's a pressing need.


theferrit32:
👍 agree with minimizing custom setup stuff

theferrit32 (approved):

PR #542, opened 2025-04-09
Fix allele translator to_spdi issue for RLE deletion

theferrit32:


src/ga4gh/vrs/extras/translator.py
```diff
@@ -9,16 +9,29 @@
...
+    """Protocol for translating VRS objects to other string expressions
+
+    This protocol defines a callable interface for translating a variation
+    string into a VRS object, with optional keyword arguments for customization.
+    """
+
+    def __call__(self, var: str, **kwargs) -> models._VariationBase | None:
+        """Translate variation `var` to VRS object"""
+
+
```
korikuzma, line 34: Should it be this?

```suggestion
class VariationToProtocol(Protocol):
    """Protocol for translating VRS objects to other string expressions

    This protocol defines a callable interface for translating a VRS object
    into variation strings, with optional keyword arguments for customization.
    """

    def __call__(self, vo: models._VariationBase, **kwargs) -> list[str]:
        """Translate vrs object `vo` to variation string expressions"""
```

src/ga4gh/vrs/extras/translator.py
```diff
@@ -110,10 +123,15 @@ def translate_from(
         msg = f"Unable to parse data as {', '.join(formats)}"
         raise ValueError(msg)
 
-    def translate_to(self, vo: models._VariationBase, fmt: str) -> str:
-        """Translate vrs object `vo` to named format `fmt`"""
+    def translate_to(self, vo: models._VariationBase, fmt: str, **kwargs) -> str:
```
korikuzma, line 126: I think our return type is wrong

```suggestion
    def translate_to(self, vo: models._VariationBase, fmt: str, **kwargs) -> list[str]:
```

src/ga4gh/vrs/extras/translator.py
```diff
@@ -410,10 +438,44 @@ def _to_spdi(
         """
         sequence = f"ga4gh:{vo.location.get_refget_accession()}"
         aliases = self.data_proxy.translate_sequence_identifier(sequence, namespace)
-        aliases = [a.split(":")[1] for a in aliases]
+        aliases = [a.split(":")[1] for a in aliases]  # TODO why split here?
```
korikuzma, line 441: An example of an alias that is returned is `refseq:NC_000021.9`, in the final SPDI expression we don't care about the curie prefix (`refseq:`)

```suggestion
        aliases = [a.split(":")[1] for a in aliases]
```

src/ga4gh/vrs/normalize.py
```diff
@@ -221,6 +221,33 @@ def _normalize_allele(input_allele, data_proxy, rle_seq_limit=50):
     return new_allele
 
 
+def denormalize_reference_length_expression(
```
korikuzma, line 224: Thank you for adding this!!!

korikuzma (commented):
Great work! Made some comments, let me know if I'm misunderstanding!

src/ga4gh/vrs/extras/translator.py
korikuzma: We might also want to consider a more descriptive name than `VariationToProtocol`, maybe like `VrsToStrFmtProtocol` or something similar?

theferrit32: @korikuzma I like the idea of adding `Str` into the class name. I'll also add a `From` protocol

korikuzma (approved):

PR #543, opened 2025-04-10
feat: define VRS schema version in package

jsstevenson:
supercedes #467. In that PR, I defined a hook for setuptools that would use the tag info in a submodule and auto-update a variable with the latest VRS schema tag, so that wheels built by our release workflow would contain the right value. There were probably going to be some hiccups that way, though, and I also couldn't get it to be testable in github actions on account of how actions/checkout seemed to be pulling the git repos.

Here, we just define a little old constant with the schema version. This is a fully manual process, but theoretically we won't need to update it that often, so it should be fine.


jsstevenson:
I was thinking about how to test this. It doesn't make sense to me to fail if the values are inequal but warnings in GitHub actions are fairly obscure so I'm not sure if this is helpful at all.

```
      - run: |
          VRS_VERSION=$(python3 -c "from ga4gh.vrs import VRS_VERSION; print(VRS_VERSION)")
          echo "vrs_version=$VRS_VERSION" >> $GITHUB_OUTPUT

      - run: |
          response=$(curl -s "https://api.github.com/repos/ga4gh/vrs/releases")
          latest_tag=$(echo "$response" | jq -r '.[0].tag_name')
          echo "latest_tag=$latest_tag" >> $GITHUB_OUTPUT

      - name: Compare versions and warn if mismatched
        run: |
          INSTALLED_VERSION="${{ steps.installed_version.outputs.vrs_version }}"
          LATEST_VERSION="${{ steps.latest_release.outputs.latest_tag }}"
          echo "Installed package version: $INSTALLED_VERSION"
          echo "Latest release tag: $LATEST_VERSION"
          if [ "$INSTALLED_VERSION" != "$LATEST_VERSION" ]; then
            # Emit a warning that will be visible in the GitHub Actions log and UI.
            echo "::warning file=version_check.yml,line=1,col=1::Warning: Installed version ($INSTALLED_VERSION) does not match latest release ($LATEST_VERSION)!"
          fi
```

korikuzma (approved):

PR #545, opened 2025-04-11
feat: add optional suppression of allele collection

jsstevenson:
* VCF annotator should not collect alleles if no output is requested by user. This has major implications for memory usage, apparently.
* Provide this in a relatively OOP-friendly way without producing a breaking change

At the next major version, there is a cleaner way to do this and it should be implemented at that point


quinnwai:
Thanks for the PR! This would definitely help address my OOM use case on Terra. For general context:
- Ran vrs-annotator workflow on Terra w/ 16GB memory allocated
- Workflow terminated unexpectedly for file input ~4GB at consistently around the 2 hour mark

original slack thread [here](https://the-anvil.slack.com/archives/C070TJDJNBF/p1744396636729539)

jsstevenson:
Yeah wrote up some brain dump here https://github.com/ga4gh/vrs-python/issues/546

PR #556, opened 2025-06-24
feat: enable recomputation of existing annotations in input VCFs

jsstevenson:


jsstevenson:
not going this direction anymore but this is a good reminder of how to use the pysam/bcftools api

PR #559, opened 2025-06-27
fix: use correct error message for ref vs actual ref

jsstevenson:
Original message is a little ambiguous -- edited for clarity

eg

```python
"Reference mismatch at GRCh38:7 position 140753335-140753336 (input gave 'G' but correct ref is 'C')
```


jsstevenson:
Everything's passing with the exception of #560

src/ga4gh/vrs/dataproxy.py
korikuzma, line 1: I _think_ the original was correct, but maybe we could just revise the warning message? 

Like "Input expression expected {ref} but the actual sequence on this position is {expected_ref}"?

jsstevenson: Yes, this PR is just modifying the message - it wasn't throwing an error incorrectly, the description of the error was backwards

korikuzma (approved):
I like this!

PR #569, opened 2025-07-23
Remove vcf annotator healthcheck

### LINKED ISSUE #530: VCF annotator should not check for GRCh38
This check is only true in the case where seqrepo has been built with that specific assembly

https://github.com/ga4gh/vrs-python/blob/main/src/ga4gh/vrs/extras/annotator/cli.py#L178

theferrit32:
Close #530

### DIFF
--- a/src/ga4gh/vrs/dataproxy.py
+++ b/src/ga4gh/vrs/dataproxy.py
@@ -247,16 +247,17 @@ class SeqRepoRESTDataProxy(_SeqRepoDataProxyBase):
 
     rest_version = "1"
 
-    def __init__(self, base_url: str):
+    def __init__(self, base_url: str, disable_healthcheck: bool = False):
         """Initialize REST-based dataproxy instance.
 
         :param base_url: root URL to server
         """
         super().__init__()
         self.base_url = f"{base_url}/{self.rest_version}/"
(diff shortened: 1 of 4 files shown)


larrybabb:
@jsstevenson would you please review this. the think-tank group decided the healthcheck() was not necessary so we removed it. If you have a counter arg, please make it, otherwise we will move forward.

jsstevenson:
I'd be fine with removal for now. The bigger problem is that the VCF annotator suppresses all exceptions and just dumps them verbatim onto the VCF in an `ERRORS` info field. I think in reality we should be more specific about exception handling, such that some things get recognized as annotation-related errors (like coordinate wonkiness or something), and other things should just interrupt annotation entirely (like a SeqRepoRestDataProxy HTTP error). This was a bandaid for the latter.

jsstevenson (approved):

theferrit32:
After discussion we decided that at least the filesystem and the REST based dataproxies should work the same way, in that they will fail at construction time if the underlying data source is not accessible. This latest commit adds a call to the `/ping` endpoint in the `__init__` function of the REST dataproxy and throws an exception if it fails

theferrit32:
I have no idea why the tests are failing because they run locally with I do `pytest --vcr-record=none`

PR #576, opened 2025-08-13
fix(notebooks): update notebooks to use vrs 2.x

### LINKED ISSUE #575: Error in CNVtranslator notebook example
The [CNVtranslator notebook](https://github.com/ga4gh/vrs-python/blob/main/notebooks/getting_started/5_Exploring_the_CnvTranslator.ipynb) no. 6 uses an EFO class (CURIE) as input[^1]. However, the current translator fails (requires the string label value).

```python
cnc = cnv_translator.translate_from("NC_000014.9:g.45002867_45015056del", "hgvs", copy_change="EFO:0030067")
```
```
...
pydantic_core._pydantic_core.ValidationError: 1 validation error for CopyNumberChange
copyChange
  Input should be 'complete genomic loss', 'high-level loss', 'low-level loss', 'loss', 'regional base ploidy', 'gain', 'low-level gain' or 'high-level gain' [type=enum, input_value='EFO:0030067', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/enum
```
This works:
```python
cnc = cnv_translator.translate_from("NC_000014.9:g.45002867_45015056del", "hgvs", cop\
y_change="loss")
```

This is probably due to a change in the enum values from the well formed curies to malformed string labels. This is a nice example why this shouldn't be done: `EFO:0030067` is variously labeled as "loss" (as used in VRS v2) or [(authoritative) "copy number loss"](http://www.ebi.ac.uk/efo/EFO_0030067). I am myself not consistent in the use since it is a _label_.

Recommendation: Use (or at least allow) the CURIEs. And/or use the exact labels from the ontology definition.

korikuzma:
close #575 

* copy_change use free text string instead of efo code curie
* update field names in sequence reference

### DIFF
--- a/notebooks/getting_started/5_Exploring_the_CnvTranslator.ipynb
+++ b/notebooks/getting_started/5_Exploring_the_CnvTranslator.ipynb
@@ -169,9 +169,9 @@
     {
      "data": {
       "text/plain": [
-       "{'id': 'ga4gh:CX.XQt04FoCIptvgp6GtE2qjEaUJC7cr1wo',\n",
+       "{'id': 'ga4gh:CX.SlJJkuXM8b0b63ADyDMxV-MpLBi7BhfI',\n",
        " 'type': 'CopyNumberChange',\n",
-       " 'digest': 'XQt04FoCIptvgp6GtE2qjEaUJC7cr1wo',\n",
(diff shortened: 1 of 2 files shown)


korikuzma:
@ga4gh/vrs-python-maintainers could this get a review?

PR #578, opened 2025-08-15
feat: add deprecation warning for beacon translation

### LINKED ISSUE #568: Rename & re-engineer misleading "beacon" Translator format
The `_Translator` class defines a `beacon` format which has no known relationship to the Beacon standard but apparently mimics a non-codified use in the _query_ fields of some _front-end_ implementations (but does not conform to any Beacon definition):

```
r"(?P<chr>[^-]+)\s*:\s*(?P<pos>\d+)\s*(?P<ref>\w+)\s*>\s*(?P<alt>\w+)"
```

The Beacon specification allows the [definition of genomic variations](https://github.com/ga4gh-beacon/beacon-v2/blob/c6558bf2e6494df3905f7b2df66e903dfe509500/models/src/beacon-v2-default-model/genomicVariations/defaultSchema.yaml#L11) as

```
      - $ref: https://w3id.org/ga4gh/schema/vrs/1.3/vrs.json#/definitions/MolecularVariation
      - $ref: https://w3id.org/ga4gh/schema/vrs/1.3/vrs.json#/definitions/SystemicVariation
      - $ref: '#/$defs/LegacyVariation'
```

... where a `LegacyVariation` is an object with properties `location` (_i.e._ `.../vrs/1.3/vrs.json#/definitions/Location`), `variantType`, `referenceBases` and `alternateBases`.

#### Recommendation

* rename the misleading `beacon` translator to `beacon-legacy`
* define an appropriate input format, preferably as an object using the Beacon v2 definition (or a stringified/list version of this) and integrate it with the appropriate method
* coordinate with/ the Beacon developers :-)

korikuzma:
close #568

* One of the Beacon leads (@mbaudis) requested this to be removed since it does not follow the standard, for now we will add a deprecation message and remove this in v3.
* Added `typing_extensions` as an `extras` dep

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
     "requests",
     "canonicaljson",
     "setuptools>=78.1.0", # TODO remove this pin caused by HGVS issue
+    "typing-extensions",
 ]
 
 [project.optional-dependencies]
--- a/src/ga4gh/core/models.py
+++ b/src/ga4gh/core/models.py
@@ -14,6 +14,7 @@ from pydantic import (
     StringConstraints,
     model_validator,
 )
+from typing_extensions import Self
 
 from ga4gh.core.identifiers import GA4GH_IR_REGEXP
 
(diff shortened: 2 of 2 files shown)


ahwagner:
For those that don't want to do major version upgrades–perhaps a related PR that just adds a deprecation warning for the 2.x series? Just not sure when we want to pull the trigger on a major version release, and if there are other features we might want to merge in to something like that.

jsstevenson:
Maybe we could merge an implementation of a deprecation warning now, and merge this PR into a 3.x branch? I think we'd want to assess whether there are other breaking changes we might want to include in a new major version release before we put one out

korikuzma:
@ga4gh/vrs-python-maintainers This is ready for review now

src/ga4gh/vrs/extras/translator.py
```diff
@@ -11,6 +11,8 @@
 from collections.abc import Mapping
 from typing import Protocol
 
+from typing_extensions import deprecated
```
jsstevenson, line 14: if we're importing it by name then it should probably be added as an explicit dependency to pyproject.toml

jsstevenson (changes requested):

PR #591, opened 2025-11-16
Add vcr_config to conftest to filter headers and auto-decompress resp…

### LINKED ISSUE #588: Ignore and exclude non-useful fields in VCR cassettes
Headers:

Request:
- User-Agent

Response
- Date
- Server
- User-Agent

Maybe just exclude all request and response headers? We are comparing the URI and params and content for equality. Don't need anything else...? If the same request to the same URI with the same params and content returns the same content, the test should pass.

https://github.com/biocommons/hgvs/pull/805/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128

theferrit32:
…onses. Refresh and clean up all cassettes. Remove all_played checks.

Close #588


jsstevenson (approved):
Motivation behind removing `all_played`?

PR #614, opened 2026-02-12
Treat all empty/none as '.' in string fields even though pysam does not

### LINKED ISSUE #610: VCF annotator should emit `.` for `missing` values in string-typed INFO fields
Labeling as a bug though it's not 100% clear that it is. `pysam` converts numeric types to the missing value `.` but not for string types. The VCF spec seems to suggest that is incorrect and `.` should be used. I removed the old manual mapping of `None` values to `.` because I assumed pysam would do it automatically for all the fields.

e.g. examples of INFO fields which may be written to output VCF files using the current VCF annotator in `2.2.0`:
```
VRS_States=,AT
VRS_Lengths=1,.
```

This should be:
```
VRS_States=.,AT
VRS_Lengths=1,.
```

theferrit32:
Close #610


korikuzma (approved):

korikuzma:
@theferrit32 Can we merge this?

PR #616, opened 2026-02-24
ci: remove get_branch job for pypi release

korikuzma:
* This is from over 4 years ago and I don't think we actually need this. There are times when we want to create releases off other branches (and its easy to forget to update). I think we should just remove

Let me know if I'm missing something?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,25 +4,9 @@ on:
     tags:
       - "*.*.**"
 jobs:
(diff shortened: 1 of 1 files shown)


jsstevenson (approved):

jsstevenson:
Yeah I think that's everything. I'd also be ok with triggering release from `release: create` rather than based on tags, but it ends up being basically the same so ndb

Issue #618, opened 2026-03-03
deprecated use of `model_validator(mode="after")` in `MappableConcept`


jsstevenson:
```
.venv/lib/python3.12/site-packages/variation/schemas/normalize_response_schema.py:52
  /Users/jss009/code/metakb/server/.venv/lib/python3.12/site-packages/variation/schemas/normalize_response_schema.py:52: PydanticDeprecatedSince212: Using `@model_validator` with mode='after' on a classmethod is deprecated. Instead, use an instance method. See the documentation at https://docs.pydantic.dev/2.12/concepts/validators/#model-after-validator. Deprecated in Pydantic V2.12 to be removed in V3.0.
    @model_validator(mode="after")
```

https://github.com/ga4gh/vrs-python/blob/b573565877bc56467f0afe3cd903cedd8bde1532/src/ga4gh/core/models.py#L229

korikuzma:
I just ran tests using main (b573565) and I do not see this morning. I think this was added in the wrong repo (variation normalizer issue https://github.com/cancervariants/variation-normalization/issues/653)

PR #619, opened 2026-03-24
build!: Split dev deps into dev, tests, lint, docs groups

### LINKED ISSUE #608: Split up `dev` dependency group
Currently the `dev` optional dependency group is:

https://github.com/ga4gh/vrs-python/blob/4303546fa33f15ff25824747dd6108260f209688/pyproject.toml#L56-L69

We should consider breaking this out into `tests`, `dev` and `docs`:

```toml
dev = [
    "pre-commit>=4.0.1",
    "ruff==0.9.4",
]
tests = [
    "pytest",
    "pytest-cov",
    "pytest-vcr",
    "vcrpy>=7.0.0",
    "pyyaml",
]
docs = [
    "sphinx",
    "sphinx_rtd_theme",
]
```

knQzx:
split the dev dependency group into test, lint, docs so CI jobs can install only what they need. the dev meta-group still pulls everything in so existing workflows don't break

model validator fix moved to #620 as requested

fixes #608
I emailed you about GSoC 2026 (knqzx0@gmail.com)


korikuzma (changes requested):
Thank you for this! Would you be able to split this out into two separate PRs - 1 for #608 and 1 for for #618? This will make the commit history clean. 

You can keep this one open for whichever you prefer.

knQzx:
> Thank you for this! Would you be able to split this out into two separate PRs - 1 for #608 and 1 for for #618? This will make the commit history clean.
> 
> You can keep this one open for whichever you prefer.

Yep, did it, thanks for the feedback!

pyproject.toml
```diff
@@ -53,20 +53,24 @@ extras = [
...
-    # tests
+test = [
     "pytest",
     "pytest-cov",
     "pytest-vcr",
     "vcrpy>=7.0.0",
     "pyyaml",
-    # style
+]
+lint = [
```
korikuzma, line 63: For consistency in other projects we maintain, I think we want to use `dev` here
```suggestion
dev = [
```

pyproject.toml
```diff
@@ -53,20 +53,24 @@ extras = [
...
     "ruff==0.9.4",
-    # docs
+]
+docs = [
     "sphinx",
     "sphinx_rtd_theme",
 ]
+dev = [
+    "ga4gh.vrs[test,lint,docs]",
+]
```
korikuzma:
```suggestion
```

pyproject.toml
```diff
@@ -53,20 +53,24 @@ extras = [
     "click",
     "pysam==0.23.0",  # pinned pending https://github.com/ga4gh/vrs-python/issues/560
 ]
-dev = [
-    # tests
+test = [
```
korikuzma: Apologies, was looking at conventions we use and we go with plural
```suggestion
tests = [
```

pyproject.toml
korikuzma: We use Makefile for this (so this should be updated to include these new groups):
https://github.com/ga4gh/vrs-python/blob/b573565877bc56467f0afe3cd903cedd8bde1532/Makefile#L48

korikuzma (changes requested):
Thank you for submitting this PR! There are few minor things to be changed and added. 

We should update this to use `tests` instead of `dev` https://github.com/ga4gh/vrs-python/blob/b573565877bc56467f0afe3cd903cedd8bde1532/.github/workflows/python-cqa.yml#L46

knQzx:
thanks for the review! renamed to tests, removed the dev meta-group, updated the ci workflow and makefile to use the new groups

korikuzma (approved):
Thank you!

korikuzma:
I think I will change the base branch to be for v3 since this is technically a breaking change.

PR #621, opened 2026-03-26
Fix copies=0 producing wrong VRS object type

knQzx:
copies=0 in _from_hgvs falls through to CopyNumberChange branch
because `if copies:` treats 0 as falsy. changed to `if copies is not None:`


knQzx:
found this while reading the translator - copies=0 is valid (homozygous deletion) but gets skipped since 0 is falsy in python

korikuzma (changes requested):
Great catch! Would you be able to add a test for this issue in `test_cnv_translator.py` ?

knQzx:
thanks for the feedback! added a test for it - checks that copies=0 gives back CopyNumberCount and not CopyNumberChange

korikuzma (approved):
Awesome! 🚀

PR #630, opened 2026-04-17
Fix CnvTranslator c to n

### LINKED ISSUE #629: CnvTranslator: c. HGVS inputs should be converted to n. before building SequenceLocation
# CnvTranslator: c. HGVS inputs should be converted to n. before building SequenceLocation

## Summary

`CnvTranslator._from_hgvs` currently builds a `SequenceLocation` directly from `sv.posedit.pos.{start,end}.base` for any supported HGVS input, including CDS-relative (`c.`) expressions. Per the VRS 2.0 specification, `SequenceLocation.start`/`end` are indices into the sequence identified by `refgetAccession` in that sequence's native frame — so when the refget accession points at a transcript (e.g. `NM_*`), the coordinates need to be transcript-relative (`n.`), not CDS-relative.

From [VRS 2.0 — Implied Sequence Coordinates](https://vrs.ga4gh.org/en/latest/concepts/LocationAndReference/SequenceLocation.html#implied-sequence-coordinates):

> The Sequence Location class refers to coordinates on a SequenceReference; if that sequence represents a coding transcript, then the coordinates refer to the coding transcript, and not a chromosome sequence to which it aligns.

`AlleleTranslator`'s `extract_allele_values` already handles this with a `c_to_n` step in `HgvsTools`. `CnvTranslator` does not, so a `c.` CNV input yields a `SequenceLocation` whose coordinates land in the 5' UTR instead of the intended CDS region — and whose `ga4gh:SL.*` digest differs from the correct value.

## Example

For `NM_001331029.1:c.100_200del` (CDS positions 100–200 of a transcript with a 150-base 5' UTR):

| | `start` | `end` | Location digest |
|---|---|---|---|
| Current behavior | 99 | 200 | `ga4gh:SL.D2epDHnPaOFt-BF2jWNk9c9hHFTQ-dFt` (UTR region) |
| Expected | 249 | 350 | `ga4gh:SL.K0g9SmC4z2-ayJsTHFHuIAygGNv1UMgi` (CDS region) |

## Fix

Mirror the Allele path: after the existing `is_intronic` guard and `derive_refget_accession` lookup, run `c_to_n` when `sv.type == "c"`. Regression test added using `NM_001331029.1:c.100_200del` against the expected transcript-relative output.

All existing `g.` CNV tests are unaffected (`sv.type != "c"`, so the new branch doesn't fire). Consumers passing `c.` HGVS to `CnvTranslator` will see their output `SequenceLocation` coordinates and digests update to the transcript-relative values required by the spec.

theferrit32:
Close #629

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -496,6 +496,10 @@ class CnvTranslator(_Translator):
         if not refget_accession:
             return None
 
+        # translate coding coordinates to positional coordinates, if necessary
+        if sv.type == "c":
+            sv = self.hgvs_tools.c_to_n(sv)
+
         location = models.SequenceLocation(
             sequenceReference=models.SequenceReference(
                 refgetAccession=refget_accession
--- a/tests/extras/cassettes/test_from_hgvs_coding_coordinates.yaml
+++ b/tests/extras/cassettes/test_from_hgvs_coding_coordinates.yaml
@@ -1,0 +1,39 @@
+interactions:
+- request:
+    body: null
+    headers: {}
+    method: GET
+    uri: http://localhost:5000/seqrepo/1/metadata/refseq:NM_001331029.1
+  response:
+    body:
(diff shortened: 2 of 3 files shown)


korikuzma (approved):
What a duo

PR #631, opened 2026-04-22
feat: Translate to gnomad

korikuzma:
Creating a draft PR so this doesn't get left behind. I think we just need to add some test cases? @hanars anything else that needs to be done?

### DIFF
--- a/src/ga4gh/vrs/extras/translator.py
+++ b/src/ga4gh/vrs/extras/translator.py
@@ -435,6 +435,12 @@ class AlleleTranslator(_Translator):
 
         If no alias translations are available, an empty list is returned.
 
+        If `namespace` is not None, returns gnomAD strings for the
+        specified namespace if applicable, or throws ValueError if not.
+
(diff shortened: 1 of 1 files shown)


src/ga4gh/vrs/extras/translator.py
```diff
@@ -450,12 +477,20 @@ def _to_spdi(
         SPDI and VRS use identical normalization. The incoming Allele
         is expected to be normalized per VRS spec.
         """
+        ref_seq_limit = kwargs.get("ref_seq_limit", 0)
+        return self._to_location_expression(
+            "{alias}:{start}:{ref_seq}:{alt_seq}", vo, namespace, ref_seq_limi=ref_seq_limit,
```
jsstevenson: typo here?


```suggestion
            "{alias}:{start}:{ref_seq}:{alt_seq}", vo, namespace, ref_seq_limit=ref_seq_limit,
```

